PR #12, opened 2019-06-25
feat: added vault as new lender

manuelpolzhofer:
- Vault implements the desk interface
- vault implements a simple lender


lucasvo:
I think it's a better idea to have the Vault not implement desk but be one of the "Lender Adapter". The reason for that is that in the future we want to support multiple lenders of which one should be the above adapter. From a business perspective, the entity providing the currency should get CVT token for funding it the same way we give CVT to the CDP as this CVT ultimately would be the way to claim money from the SPV in case it needs to be shut down. This would keep it much closer to how it also works with Maker. What do you think of this?

manuelpolzhofer:
I agree. I will keep the vault modul for other use cases

src/test/collateral.t.sol
```diff
@@ -263,19 +263,20 @@ contract CollateralTest is DSTest {
...
-    }
-
-    function test_permit() public {
-        assertEq(token.nonces(cal),0);
-        assertEq(token.allowance(cal, del),0);
-        token.permit(cal, del, 0, 0, true, v, r, s);
-        assertEq(token.allowance(cal, del),uint(-1));
-        assertEq(token.nonces(cal),1);
-    }
+//
```
lucasvo: @manuelpolzhofer could you remove this if it's not used?

PR #16, opened 2019-06-28
feat: added backer as a new lender adapter

manuelpolzhofer:


src/backer.sol
```diff
@@ -0,0 +1,56 @@
+pragma solidity >=0.4.24;
+
+contract LenderTokenLike {
```
lucasvo: Isn't this enough:

```suggestion
contract TokenLike {
```

.gitmodules
```diff
@@ -13,3 +13,6 @@
 [submodule "lib/dss-add-ilk-spell"]
 	path = lib/dss-add-ilk-spell
 	url = https://github.com/makerdao/dss-add-ilk-spell
+[submodule "lib/ds-token"]
```
lucasvo: Seems to be unused?

src/backer.sol
```diff
@@ -0,0 +1,56 @@
...
+    function provide(address usrC, address usrT, uint wadC, uint wadT) auth public {
+        require(tkn.balanceOf(backer)>= wadT);
+        collateral.transferFrom(usrC, backer, wadC);
+        tkn.transferFrom(backer,usrT, wadT);
+
+    }
+
+    function release(address usrC, address usrT, uint wadC, uint wadT) auth  public {
+        require(collateral.balanceOf(backer)>= wadC);
+        tkn.transferFrom(usrT,backer, wadT);
```
lucasvo:
```suggestion
        tkn.transferFrom(usrT, backer, wadT);
```

src/backer.sol
```diff
@@ -0,0 +1,56 @@
...
+    mapping (address => uint) public wards;
+    function rely(address usr) public auth { wards[usr] = 1; }
+    function deny(address usr) public auth { wards[usr] = 0; }
+    modifier auth { require(wards[msg.sender] == 1); _; }
+
+    // --- Data ---
+    LenderTokenLike public tkn;
+    LenderTokenLike public collateral;
+
+    address private backer;
```
lucasvo: why `private`?

src/backer.sol
```diff
@@ -0,0 +1,56 @@
...
+
+    }
+
+    function release(address usrC, address usrT, uint wadC, uint wadT) auth  public {
+        require(collateral.balanceOf(backer)>= wadC);
+        tkn.transferFrom(usrT,backer, wadT);
+        collateral.transferFrom(backer, usrC, wadC);
+    }
+
+    function setBacker(address usr) auth public {
```
lucasvo: I'd like to use: 
```suggestion
    function file(address usr) auth public {
```

src/backer.sol
```diff
@@ -0,0 +1,56 @@
...
+    function deny(address usr) public auth { wards[usr] = 0; }
+    modifier auth { require(wards[msg.sender] == 1); _; }
+
+    // --- Data ---
+    LenderTokenLike public tkn;
+    LenderTokenLike public collateral;
+
+    address private backer;
+
+    constructor (address tkn_, address collateral_, address backer_) public {
```
lucasvo: I think it's better not to require backer_ at instantiation. It makes the deploy more complicated. 
```suggestion
    constructor (address tkn_, address collateral_) public {
```

src/backer.sol
```diff
@@ -0,0 +1,56 @@
...
+    function transferFrom(address, address, uint) public;
+    function mint(address, uint) public;
+    function approve(address usr, uint wad) public returns (bool);
+    function balanceOf(address usr) public returns (uint);
+}
+
+contract BackerFab {
+    function deploy(address tkn_, address collateral_,address backer, address lightswitch_) public returns (address) {
+        Backer lender = new Backer(tkn_, collateral_,backer );
+        lender.rely(msg.sender);
```
lucasvo:
```suggestion
        lender.rely(msg.sender);
        lender.deny(this);
```

src/backer.sol
```diff
@@ -0,0 +1,56 @@
...
+        tkn = LenderTokenLike(tkn_);
+        collateral = LenderTokenLike(collateral_);
+        backer = backer_;
+    }
+
+    // --- Backer Methods ---
+    function provide(address usrC, address usrT, uint wadC, uint wadT) auth public {
+        require(tkn.balanceOf(backer)>= wadT);
+        collateral.transferFrom(usrC, backer, wadC);
+        tkn.transferFrom(backer,usrT, wadT);
```
lucasvo:
```suggestion
        tkn.transferFrom(backer, usrT, wadT);
```

src/backer.sol
manuelpolzhofer: I know ever private variable is publicly visible but I thought it could be a bit more secure if it is not on chain accessible.

lucasvo: What would be more secure? It's less secure because you can't see who the current backer is.

src/backer.sol
lucasvo: Also, let's use ds-note here and log the event.

PR #17, opened 2019-07-03
fix: pile chi calculation

manuelpolzhofer:


src/pile.sol
```diff
@@ -74,7 +74,9 @@ contract Pile is DSNote {
     }
     
     function file(uint fee, uint speed_) public auth note {
-        fees[fee].speed = speed_; 
+        fees[fee].speed = speed_;
+        fees[fee].chi = ONE;
+        drip(fee);
```
lucasvo: this drip should be called before changing the speed IMO.

src/test/pile.t.sol
```diff
@@ -20,13 +20,20 @@ import "ds-test/test.sol";
...
 contract PileTest is DSTest {
     Pile pile;
     TokenMock tkn;
 
-    function setUp() public {
+    Hevm hevm;
 
+    function setUp() public {
+        hevm = Hevm(0x7109709ECfa91a80626fF3989D68f67F5b1DD12D);
+        hevm.warp(0);
```
lucasvo: Can you test with something other than warp(0)? Check this out: https://github.com/makerdao/dss/pull/45/files#diff-ba2ea047fd34da8c18c72fcc967b6e4bL115 (setting it to 0 was the problem why we didn't find the rho issue)

src/test/pile.t.sol
```diff
@@ -58,7 +64,7 @@ contract PileTest is DSTest {
 
         pile.withdraw(loan,wad,address(this));
 
-        assertEq(totalBalance-wad, pile.Balance());
+         assertEq(totalBalance-wad, pile.Balance());
```
lucasvo: spacing

manuelpolzhofer:
removed `hvem.warp(0)` from tests

src/pile.sol
manuelpolzhofer: Both ways work.

PR #19, opened 2019-07-04
fix: pile drip

manuelpolzhofer:
implements #14

### DIFF
--- a/src/pile.sol
+++ b/src/pile.sol
@@ -129,15 +129,21 @@ contract Pile is DSNote {
         // --- Fee Accumulation ---
     function drip(uint fee) public {
         uint48 rho = fees[fee].rho;
+        require(now >= rho);
         uint speed = fees[fee].speed;
+        if (speed == 0) {
+            return;
+        }
         uint chi = fees[fee].chi;
         uint debt = fees[fee].debt;
-        require(now >= rho);
-        uint chi_ = sub(rmul(rpow(speed, now - rho, ONE), chi), chi);
-        uint wad = mul(debt, chi_);
-        add(Debt, wad);
+
+        uint latest = rmul(rpow(speed, now - rho, ONE), chi);
+        uint chi_ = rdiv(latest, chi);
+        uint wad = rmul(debt, chi_)-debt;
+
+        Debt = add(Debt, wad);
         fees[fee].debt = add(debt, wad);
-        fees[fee].chi = add(chi, chi_);
+        fees[fee].chi = latest;
         fees[fee].rho = uint48(now);
     }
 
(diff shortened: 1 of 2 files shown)


src/pile.sol
```diff
@@ -129,15 +129,21 @@ contract Pile is DSNote {
         // --- Fee Accumulation ---
     function drip(uint fee) public {
         uint48 rho = fees[fee].rho;
+        require(now >= rho);
         uint speed = fees[fee].speed;
+        if (speed == 0) {
```
lucasvo: Does it breafk if we not have this check? It will be more easily verifiable this way.

src/pile.sol
```diff
@@ -129,15 +129,21 @@ contract Pile is DSNote {
...
         uint chi = fees[fee].chi;
         uint debt = fees[fee].debt;
-        require(now >= rho);
-        uint chi_ = sub(rmul(rpow(speed, now - rho, ONE), chi), chi);
-        uint wad = mul(debt, chi_);
-        add(Debt, wad);
+
+        uint latest = rmul(rpow(speed, now - rho, ONE), chi);
+        uint chi_ = rdiv(latest, chi);
+        uint wad = rmul(debt, chi_)-debt;
```
lucasvo: I don't think this should be rmul but regular mul.

manuelpolzhofer: It needs to be `rmul`. I just tested it.

src/pile.sol
manuelpolzhofer: Currently, it would break. If the `fee` is not set for a loan. but I think I can avoid it.

lucasvo: Does it produce bad values or does it divide by zero? I think a revert here is fine if it's not set.

manuelpolzhofer: The admit method will get a new parameter `fee` afterwards `speed==0` won't happen anymore.

src/pile.sol
```diff
@@ -61,6 +61,8 @@ contract Pile is DSNote {
     constructor(address tkn_) public {
         wards[msg.sender] = 1;
         tkn = TokenLike(tkn_);
+        fees[0].chi = ONE;
+        fees[0].speed = ONE;
```
manuelpolzhofer: @lucasvo solves the problem

src/pile.sol
lucasvo: I'd prefer to leave it as a separate method and add a call to the reception that does both at the same time.

manuelpolzhofer: The reception is only called by the borrower. I think it makes sense to pass it in the admit method and the admit contract calls `pile.file` afterwards.

PR #20, opened 2019-07-06
test: changed systems tests to work with backer and simple adapter

manuelpolzhofer:


src/test/integration/backer.t.sol
```diff
@@ -0,0 +1,46 @@
...
+import "ds-test/test.sol";
+
+import "./system.t.sol";
+import "../../backer.sol";
+
+contract BackerLike {
+    function file(address usr) public;
+}
+
+contract BackerTest is DSTest, SystemTest {
```
manuelpolzhofer: all tests in systemTest will be executed with the backer adapter

src/test/simple/lender.sol
```diff
@@ -25,10 +25,14 @@ contract SimpleLender {
     LenderTokenLike public tkn;
     LenderTokenLike public collateral;
 
+
+    address public backer;
```
manuelpolzhofer: In the simple adapter, the backer is the contract itself.

src/deployer.sol
```diff
@@ -30,7 +30,9 @@ contract LenderFabLike {
 }
 
 contract LenderLike {
+    function backer() public returns(address);
```
lucasvo: I would split this into two interfaces because not all lenders implement backer()

src/deployer.sol
```diff
@@ -171,17 +173,25 @@ contract Deployer {
...
         lender.rely(address(desk));
-        
+
         desk.approve(lender_, uint(-1));
         pile.file("lender", lender_);
         desk.file("lender", lender_);
+        return lender_;
+    }
+
+    function deployLender(address currency_, address lenderfab_, address backer_) public returns(address) {
```
lucasvo: Why not use inheritance here and create two deployers?

src/test/integration/system.t.sol
```diff
@@ -83,17 +89,17 @@ contract SystemTest is DSTest {
 
     ManagerUser  manager;
     address      manager_;
-    BorrowerUser borrower;
+    User borrower;
```
lucasvo: spacing

src/test/simple/lender.sol
lucasvo: Why does the SimpleLender even need a backer?

src/test/integration/backer.t.sol
lucasvo: Why not simplelender?

src/test/simple/lender.sol
manuelpolzhofer: We are only using Simple lender for testing and I wanted to run the system test on both.

I wanted a way to check if the adapter received currency and collateral. In the simpleLender case,  the backer is the contract itself.

src/test/integration/backer.t.sol
manuelpolzhofer: system tests with SimplerLender are executed too.

src/test/integration/system.t.sol
manuelpolzhofer: I will add a solidity linter

src/test/simple/lender.sol
manuelpolzhofer: I can remove it. I could just overwrite the check method too. (to support two different way)

src/deployer.sol
manuelpolzhofer: fixed

src/test/simple/lender.sol
manuelpolzhofer: fixed

src/deployer.sol
manuelpolzhofer: Instead of two deployers I would initialize all variables of different LenderFabs in the lenderFab constructor and only have one `deploy` method without parameters.

manuelpolzhofer: I think I still need inheritance. I will fix it in the next PR

manuelpolzhofer: fixed. no inheritance needed. removed second deployLender method

src/test/integration/system.t.sol
```diff
@@ -135,11 +135,17 @@ contract SystemTest is DSTest {
 
     }
 
+
+    // lenderTokenAddr returns the address which holds the currency or collateral token for the lender
+    function lenderTokenAddr(address lender) public returns(address) {
```
lucasvo: why not just use a public method for `lender`?

PR #26, opened 2019-07-09
feat: pile repay supports currency buffer

manuelpolzhofer:
please review #20 first.

- repay supports a buffer in the repay amount to cover fee. 
- amount taken by tinlake never exceed debt

- modified integration test to work with fee


src/pile.sol
```diff
@@ -223,6 +223,11 @@ contract Pile is DSNote {
         // moves currency from usr to pile and reduces debt
         require(loans[loan].balance == 0,"before repay loan needs to be withdrawn");
         collect(loan);
+
+        if (wad > loans[loan].debt) {
```
manuelpolzhofer: @lucasvo major change here:

PR #31, opened 2019-07-12
feat: admin contract

manuelpolzhofer:
- admin contract with whitelist method

admin deployment and integration test will follow in the next PR

### DIFF
--- a/src/admin.sol
+++ b/src/admin.sol
@@ -18,6 +18,8 @@ pragma experimental ABIEncoderV2;
 
 contract AdmitLike {
     function admit (address registry, uint nft, uint principal, address usr) public returns(uint);
+    function update(uint loan, address registry, uint nft, uint principal) public;
+    function update(uint loan, uint principal) public;
 }
 
 contract AppraiserLike {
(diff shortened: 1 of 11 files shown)


src/admin.sol
```diff
@@ -0,0 +1,69 @@
...
+        appraiser = AppraiserLike(appraiser_);
+        pile = PileLike(pile_);
+    }
+
+    // -- Whitelist --
+    function whitelist(address registry, uint nft, uint principal, uint appraisal, uint fee, address usr) public auth returns(uint) {
+        uint loan = admit.admit(registry, nft, principal, usr);
+        appraiser.file(loan, appraisal);
+
+        (,,uint speed,) = pile.fees(loan);
```
lucasvo: Why do you need to check if it exists here? Can't we just always set it?

manuelpolzhofer: Currently not because we init the `fees[fee].chi = ONE;` in the file method.

lucasvo: Can you explain the scenario where we need this?

PR #33, opened 2019-07-18
feat: admin deploy

manuelpolzhofer:
please review #30 and #31 first

### DIFF
--- a/bin/deploy-core
+++ b/bin/deploy-core
@@ -23,9 +23,12 @@ PILE_FAB=$(dapp create PileFab)
 SHELF_FAB=$(dapp create ShelfFab)
 COLLATERAL_FAB=$(dapp create CollateralFab)
 DESK_FAB=$(dapp create DeskFab)
+ADMIT_FAB=$(dapp create AdmitFab)
+ADMIN_FAB=$(dapp create AdminFab)
+
 
 # Deploy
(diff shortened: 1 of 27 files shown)


src/deployer.sol
```diff
@@ -24,17 +24,17 @@ import { Pile } from "./pile.sol";
...
+import { Admin } from "./admin.sol";
 
 contract LenderFabLike {
     function deploy(address,address,address) public returns (address);
 }
 
-contract LenderLike {
+contract WardsLike {
     function rely(address) public;
     function file(address) public;
```
lucasvo: WardsLike doesn't necessarily have file()

bin/deploy-core
```diff
@@ -23,16 +23,24 @@ PILE_FAB=$(dapp create PileFab)
...
 $(seth send $DEPLOYER 'deployLightSwitch()')
 $(seth send $DEPLOYER 'deployCollateral()')
 $(seth send $DEPLOYER 'deployPile(address)' $CURRENCY)
 $(seth send $DEPLOYER 'deployShelf(address)' $APPRAISER)
 $(seth send $DEPLOYER 'deployValve()')
 $(seth send $DEPLOYER 'deployDesk()')
+$(seth send $DEPLOYER 'deployAdmit()')
+$(seth send $DEPLOYER 'deployAdmin(address)' $APPRAISER)
+
+$(seth send $APPRAISER 'rely(address)' $DEPLOYER)
```
manuelpolzhofer: The admin contract should be allowed to use the appraiser.

I think it is better if the deployer adds permission instead of `GOD` with `seth`. 

What do you think?

src/deployer.sol
manuelpolzhofer: fixed

PR #34, opened 2019-07-19
feat: backer health script

manuelpolzhofer:


bin/backer-health
```diff
@@ -0,0 +1,42 @@
...
+echo "Collateral: $COLLATERAL"
+echo "Lender: $LENDER"
+
+
+message Backer Balances
+echo "Backer Address: $BACKER_ETH_FROM"
+echo "ETH Balance: $(seth --from-wei $(seth balance $BACKER_ETH_FROM)) ETH"
+
+BALANCE=$(seth call $CURRENCY 'balanceOf(address)' $BACKER_ETH_FROM)
+echo "Backer Currency: $(seth --from-wei $(node <<< "console.log(parseInt('$BALANCE', 16))")) DAI"
```
manuelpolzhofer: @lucasvo  `seth` is not supporting hex to decimal conversion, yet.

For `printf` the number was to big.  I didn't want to spend too much time on it, so I used `node` for know.

manuelpolzhofer: `echo "ibase=16; HEXNUMBER" | bc` can't handle `0x` prefix and lowercase letters.

lucasvo: This does the trick: 
```suggestion
echo "Backer Currency: $(seth --from-wei $((16#${BALANCE:2}))) DAI"
```

manuelpolzhofer: Doesn't work either. We will keep the dependancy for now.

PR #36, opened 2019-07-22
feat: repay without defining amount + more integration tests

manuelpolzhofer:


src/reception.sol
```diff
@@ -69,5 +71,13 @@ contract Reception is TitleOwned {
         pile.repay(loan,wad,usrT);
         shelf.release(loan, usr);
         desk.balance();
-    }    
+    }
+    function repay(uint loan, address usrT, address usr) public owner(loan) {
+        pile.collect(loan);
+
+        (uint debt,,,) = pile.loans(loan);
+        pile.repay(loan, debt, usrT);
```
lucasvo: I think it would be nicer if you just called the above repay function after doing `pile.collect()`

manuelpolzhofer: fixed

PR #42, opened 2019-08-02
Refactored deploy scripts

lucasvo:


bin/deploy-backer-lender
```diff
@@ -11,33 +11,28 @@ export SOLC_FLAGS="--optimize"
...
-LENDER_FAB=$(dapp create BackerFab $BACKER_ETH_FROM)
+LENDER_FAB=$(dapp create BackerFab $BACKER_ADDR)
 
 $(seth send $DEPLOYER 'deployLender(address, address)' $CURRENCY $LENDER_FAB)
 LENDER=0x$(seth call $DEPLOYER 'lender()(address)')
 $(seth send $DESK 'file(bytes32,address)' "$(seth --to-bytes32 "$(seth --from-ascii "lender")")" $LENDER)
 $(seth send $LENDER 'rely(address)' $DESK)
 
+if [[ -n "$BACKER_ETH_FROM" ]]; then
+    seth send $CURRENCY 'approve(address,uint)' $LENDER 100000 --keystore=$BACKER_ETH_KEYSTORE --password=$BACKER_ETH_PASSWORD --from=$BACKER_ETH_FROM
```
manuelpolzhofer:
```suggestion
    seth send $CURRENCY 'approve(address,uint)' $LENDER ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff --keystore=$BACKER_ETH_KEYSTORE --password=$BACKER_ETH_PASSWORD --from=$BACKER_ETH_FROM
```

manuelpolzhofer:
```suggestion
    seth send $CURRENCY 'approve(address,uint)' $LENDER ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff --keystore=$BACKER_ETH_KEYSTORE --password=$BACKER_ETH_PASSWORD --from=$BACKER_ETH_FROM
```

manuelpolzhofer: I would move it into an additional script. 
`--keystore=$BACKER_ETH_KEYSTORE --password=$BACKER_ETH_PASSWORD ` is not always the case. If we are using a ledger we only need `$BACKER_ETH_FROM`

README.md
```diff
@@ -8,7 +8,9 @@ This is a very rough prototype and is not ready for production use.
 ### Testnet
 The easiest way to run these scripts is with `dapp testnet`. Use this command to start a geth devserver.
 
+
 ### Pre-deploy:
+All deployed addresses are written to `addresses-$(seth chain).json`. To start with a fresh deploy, remove the address file.
```
manuelpolzhofer:
```
{
  "NFT_COLLATERAL": "0x...",
  "GOD_ADDR": "0x...",
  "CURRENCY": "0x...",
```
Is needed otherwise it would fail

bin/deploy-all-backer
```diff
@@ -20,6 +22,11 @@ if [[ "$ETHERSCAN_API_KEY" != "" ]]; then
     echo "Found ETHERSCAN_API_KEY in environment, will attempt to verify contracts."
 fi
 
+loadAddresses $ADDRESS_FILE
+if [[ -z "$GOD_ADDR" || -z "$CURRENCY" ]]; then
```
manuelpolzhofer: I would require `NFT_COLLATERAL` as well. Technically it is not required for the deployment but if it is not included in the final JSON it won't work in the UI

bin/admin
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+BIN_DIR=${BIN_DIR:-$(cd "${0%/*}"&&pwd)}
+ADDRESS_DIR=${ADDRESS_DIR:-$PWD/addresses}
+ADDRESS_FILE="${ADDRESS_DIR}/addresses-$(seth chain).json"
+
+export ETH_GAS=${ETH_GAS:-"7000000"}
```
manuelpolzhofer: If not set let's use a default gas price `export ETH_GAS_PRICE=1100000000` (1.1 gwei)

README.md
lucasvo: No, those can also  be exported as environment variables, they don't need to be in the file. Why do we need NFT_COLLATERAL?

bin/deploy-all-backer
lucasvo: "Technically it is not required" - we will hopefully soon have a deployment that will have multiple. I strongly disagree that it should be included. UI and contracts should be separated.

bin/admin
lucasvo: OK, but that would have to be changed in all then.

lucasvo: done

PR #53, opened 2019-08-22
added require for appraisal and principal in whitelist

manuelpolzhofer:
it was possible to whitelist a loan with a zero appraisal value.
Which results in a loan which could never be borrowed.

The contracts should not allow whitelisting such a loan at all.


manuelpolzhofer:
dynamic appraiser in the future could have an appraisal of zero which should not allow borrowing.

Close Ticket.

PR #55, opened 2019-08-23
feat: pile improvements

manuelpolzhofer:
- added incDebt and decDebt
- removed collect from borrow
- added init loan method

### DIFF
--- a/src/pile.sol
+++ b/src/pile.sol
@@ -35,7 +35,7 @@ contract Pile is DSNote {
 
     // --- Data ---
     TokenLike public tkn;
-   
+
     // https://github.com/makerdao/dsr/blob/master/src/dsr.sol
     struct Fee {
         uint debt;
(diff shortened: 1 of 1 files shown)


src/pile.sol
```diff
@@ -190,18 +200,22 @@ contract Pile is DSNote {
...
+        Balance = add(Balance, wad);
+
+    }
+
     // borrow() creates a debt by the borrower for the specified amount. 
     function borrow(uint loan, uint wad) public auth note {
         uint fee = loans[loan].fee;
         drip(fee);
-        loans[loan].chi = fees[fee].chi;
-        collect(loan);
```
manuelpolzhofer: the `collect` call was unnecessary.

`collect` updates `loans[loan].debt` `loans[loan].chi`.

`loans[loan].chi` is already assigned in the new `initLoan` method previously in line 197.

`loans[loan].debt` always stays zero. Therefore the collect call is not needed.

PR #61, opened 2019-09-10
feat: wards for deployer contract

manuelpolzhofer:
Instead of removing the deployer as a ward from all contracts after the deployment.
Only permissioned addresses are allowed to call the deployer.

### DIFF
--- a/src/deployer.sol
+++ b/src/deployer.sol
@@ -231,17 +231,8 @@ contract Deployer {
 
     }
 
-    function resign() internal {
-        address self = address(this);
-        pile.deny(self);
-        desk.deny(self);
-        shelf.deny(self);
-        admit.deny(self);
-        collateral.deny(self);
-        valve.deny(self);
-        title.deny(self);
-        admin.deny(self);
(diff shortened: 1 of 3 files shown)


manuelpolzhofer:
@lucasvo @ilinzweilin 
I added a new `resign` method to the deployer. We also need the wards because otherwise, it would be theoretically possible that somebody calls a deployMethod again during the deployment.

lucasvo:
@manuelpolzhofer have you thought about adding a method that will call selfdestruct on the deployer? I think that would be handy

manuelpolzhofer:
true. I always forget that solidity has this nice `selfdestruct()` functionality. On the other hand with the current implementation, it would be still possible to re-active the `deployer` again in case it is needed.

lucasvo:
@manuelpolzhofer I think the selfdestruct would just be an option people could pick to be 100% certain that the deployer can't do anything anymore.

lucasvo:
(i.e. it shouldn't be be a requirement)

PR #64, opened 2019-09-17
add admin proxy contract for tinlake demo env

ilinzweilin:


ilinzweilin:
moving pr to tinlake-playground repo

PR #78, opened 2019-10-28
chore-76 remove reception

ilinzweilin:
- added loan owner check to pile & shelf 
- removed reception
- changed integration tests
- made desk.balance() public


src/desk.sol
```diff
@@ -79,7 +79,7 @@ contract Desk is Switchable {
     }
 
     // --- Desk ---
-    function balance() public auth {
+    function balance() public {
```
manuelpolzhofer: Removing the `auth` here will cause an unstable version. Which we should be aware of.

Because we currently not support async operations:

If everybody would use the proxies it is still fine.

If a borrower would use the core contracts directly, we have some edge cases
- borrower repays but never picks up his NFT
    - `pile.want()`  => negative and `valve.want()` => 0
    - `balance()` check would always fail

I will open a hackMD for the async feature.

lucasvo: @manuelpolzhofer I think that's fine. that would be a feature of the system that we can shield most people from by letting them use the proxies.

src/pile.sol
```diff
@@ -217,7 +218,7 @@ contract Pile is DSNote {
     }
 
     // repay() a certain amount of token from the user to the Pile
-    function repay(uint loan, uint wad, address usr) public auth note {
+    function repay(uint loan, uint wad, address usr) public owner(loan) note {
```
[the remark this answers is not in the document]
ilinzweilin: @manuelpolzhofer  The same applies to the withdraw method, right?

PR #79, opened 2019-10-30
File and depend methods

### LINKED ISSUE #72: consistent use and extensions of file pattern
every `file method` should have a `bytes32 what`.

Every contract which has a dependency on another contract should have `file` methods to change the dependency.

Example for `reception.sol`
```javascript
    // --- Data ---
    DeskLike desk;
    ShelfLike shelf;
    PileLike pile;

 function file(bytes32 what, address addr) public auth {
        if (what == "pile") pile = PileLike(addr);
        else if (what == "shelf") shelf = ShelfLike(addr);
        else if (what == "desk") desk = DeskLike(addr);
        else revert();
    }
```

I think we should use the file pattern only for external configuration. If one core contract calls another core contract we should use other method names.

charleenfei:
Closes #72 and #60 .

Sets up a `file` method for external configurations, and introduces `depend` for core contract dependencies.

### DIFF
--- a/src/admin.sol
+++ b/src/admin.sol
@@ -57,6 +57,7 @@ contract Admin {
     function depend (bytes32 what, address addr) public auth {
         if (what == "pile") { pile = PileLike(addr); }
         else if (what == "admit") { admit = AdmitLike(addr); }
+        else if (what == "appraiser") { appraiser = AppraiserLike(addr); }
         else revert();
     }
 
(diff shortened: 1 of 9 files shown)


src/admin.sol
```diff
@@ -67,6 +73,11 @@ contract Admin {
         return loan;
     }
 
+    function file(bytes32 what, address addr) public auth {
```
manuelpolzhofer: if we call it `depend` we should also move this method into the new `depend`.

The idea is to use `depend` for contract dependencies and `file` for other external input like a
principal or a nft etc.

charleenfei: actually, i was thinking about it as core contracts v external (substitutable) module contracts (appraiser being substituted eventually by potentially underwriter system, etc). but we could also put all contract dependencies into `depend` and `file` just for external 'data' input. not opinionated on this.

manuelpolzhofer: ok. maybe let's use `depend` for all contract dependencies for now.

PR #103, opened 2019-11-22
tiny mock new Desk (Tranche Manager)

### LINKED ISSUE #95: create mock contract for the lender side
create one mock contract for the lender side to allow async dev for borrower and lender side

charleenfei:
Closes #95


src/test/simple/lifeguard.sol
```diff
@@ -0,0 +1,84 @@
...
+contract SimpleLifeguardFab {
+    function deploy(address pile_, address operator_, address lightswitch_) public returns (address) {
+        SimpleLifeguard lifeguard = new SimpleLifeguard(pile_, operator_);
+        lifeguard.rely(msg.sender);
+        return address(lifeguard);
+    }
+}
+
+// Tranche Manager
+contract SimpleLifeguard {
```
manuelpolzhofer: why is it called `lifeguard`?

src/test/simple/lifeguard.sol
```diff
@@ -0,0 +1,84 @@
...
+
+    mapping (address => uint) public wards;
```
manuelpolzhofer: I think we don't need wards here even for the final tranche manager. Everybody could call `balance`

src/test/simple/lifeguard.sol
```diff
@@ -0,0 +1,84 @@
+pragma solidity >=0.4.24;
+
+contract TokenLike {
+    function transferFrom(address, address, uint) public;
+    function mint(address, uint) public;
+    function approve(address usr, uint wad) public returns (bool);
```
manuelpolzhofer: I think we don't need the approve here.

src/test/simple/lifeguard.sol
manuelpolzhofer: are we calling the tranche manager lifeguard?

src/test/simple/lifeguard.sol
```diff
@@ -0,0 +1,84 @@
...
+}
+
+contract OperatorLike {
+    function provide(address,address,uint,uint) public;
+    function release(address,address,uint,uint) public;
+    function free(address, uint) public;
+}
+
+contract SimpleLifeguardFab {
+    function deploy(address pile_, address operator_, address lightswitch_) public returns (address) {
```
manuelpolzhofer: I think we don't need the lightswitch here.

src/test/simple/lifeguard.sol
```diff
@@ -0,0 +1,84 @@
...
+    function transferFrom(address, address, uint) public;
+    function mint(address, uint) public;
+    function approve(address usr, uint wad) public returns (bool);
+}
+
+contract PileLike {
+    function want() public returns (int);
+}
+
+contract OperatorLike {
```
manuelpolzhofer: I thought we would use the names `give` and `take` for the new operator?

src/test/simple/lifeguard.sol
```diff
@@ -0,0 +1,84 @@
...
+        if (wadT > 0) {
+            operator.provide(address(pile), uint(wadT));
+
+        } else {
+            operator.release(address(pile), uint(wadT*-1));
+        }
+    }
+}
+
+contract SimpleOperator {
```
manuelpolzhofer: For the mock version, i would have not extra generated a simple operator but i think both is fine.

src/test/simple/lifeguard.sol
```diff
@@ -0,0 +1,84 @@
...
+    // --- Lender Side Methods ---
+    function provide(address usrT, uint wadT) public {
+        tkn.mint(usrT, wadT);
+    }
+
+    function release(address usrT, uint wadT) public {
+        tkn.transferFrom(usrT, address(this), wadT);
+    }
+
+    function free(address usr, uint wad) public {
```
manuelpolzhofer: we can remove it for the simple version

src/test/simple/lifeguard.sol
```diff
@@ -0,0 +1,84 @@
...
+
+        } else {
+            operator.release(address(pile), uint(wadT*-1));
+        }
+    }
+}
+
+contract SimpleOperator {
+
+    mapping (address => uint) public wards;
```
manuelpolzhofer: I don't think we need auth for the simple versions. Currently the `auth` is also not used.

src/test/simple/lifeguard.sol
charleenfei: we can also call it tranche manager ;) it was to go with surfer but if we are removing it then we can update.

src/test/simple/lifeguard.sol
```diff
@@ -0,0 +1,52 @@
...
+        return address(manager);
+    }
+}
+
+// Tranche Manager
+contract SimpleTrancheManager {
+
+    // --- Data ---
+    PileLike public pile;
+    // simple lifeguard = 1 tranche/1 operator for now
```
manuelpolzhofer:
```suggestion

```

src/test/simple/lifeguard.sol
```diff
@@ -0,0 +1,52 @@
...
+    function transferFrom(address, address, uint) public;
+    function mint(address, uint) public;
+}
+
+contract PileLike {
+    function want() public returns (int);
+}
+
+
+contract SimpleTrancheManagerFab {
```
manuelpolzhofer:
```suggestion
contract TrancheManagerFab {
```

src/test/simple/manager.sol
```diff
@@ -0,0 +1,50 @@
...
+        int wad = pile.want();
+        if (wad > 0) {
+            give(address(pile), uint(wad));
+
+        } else {
+            take(address(pile), uint(wad*-1));
+        }
+    }
+
+    // --- Operator Methods ---
```
manuelpolzhofer:
```suggestion
    // --- Operator Methods ---
```

Issue #86, opened 2019-11-15
Move interest calculation into a separate contract


ilinzweilin:
- remove interest calculation from the pile 
- create Interest contract
- keep loan chi values stored inside pile (could be used by other contracts)

PR #107, opened 2019-11-26
chore: #86 debt module separation

ilinzweilin:
- remove debt and interest calculation from pile into beans contract
- beans stores pie values of each loan and computes current debt for each loan
- move fee "category" management to beans
- make admin to use beans contract directly to init new fee "categories"
- adjust tests


src/beans.sol
```diff
@@ -0,0 +1,176 @@
...
+    function rely(address usr) public auth note { wards[usr] = 1; }
+    function deny(address usr) public auth note { wards[usr] = 0; }
+    modifier auth { require(wards[msg.sender] == 1); _; }
+    
+    // --- Data ---
+    // https://github.com/makerdao/dsr/blob/master/src/dsr.sol
+    struct Fee {
+        uint debt;
+        uint chi;
+        uint speed; // Accumulation per second
```
manuelpolzhofer:
```suggestion
        uint secInterest; // Accumulation per second
```

src/beans.sol
```diff
@@ -0,0 +1,176 @@
...
+
+    function rdiv(uint x, uint y) internal pure returns (uint z) {
+        z = add(mul(x, ONE), y / 2) / y;
+    }
+
+    function div(uint x, uint y) internal pure returns (uint z) {
+        z = x / y;
+    }
+
+    function initLoanDebt(uint loan, uint fee, uint wad) public auth note {
```
manuelpolzhofer: I think it is not always `init`. In case we would allow `partialBorrow`.

src/beans.sol
```diff
@@ -0,0 +1,176 @@
...
+    // --- Auth ---
+    mapping (address => uint) public wards;
+    function rely(address usr) public auth note { wards[usr] = 1; }
+    function deny(address usr) public auth note { wards[usr] = 0; }
+    modifier auth { require(wards[msg.sender] == 1); _; }
+    
+    // --- Data ---
+    // https://github.com/makerdao/dsr/blob/master/src/dsr.sol
+    struct Fee {
+        uint debt;
```
manuelpolzhofer:
```suggestion
        uint totalDebt;
```

src/beans.sol
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    function div(uint x, uint y) internal pure returns (uint z) {
+        z = x / y;
+    }
+
+    function initLoanDebt(uint loan, uint fee, uint wad) public auth note {
+        drip(fee);
+        uint pie_ = calcPie(fees[fee].chi, wad);
+        pie[loan] = pie_;
```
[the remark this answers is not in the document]
manuelpolzhofer: I think to allow partialBorrow we could already write:

`pie[loan] = add(pie[loan], calcPie(wad, fee));`

src/beans.sol
```diff
@@ -0,0 +1,176 @@
...
+            (chi, ,) = compounding(fee);
+        }
+        return calcDebt(chi, pie[loan]);
+    }
+
+    function debtOf(uint loan, uint fee) public view returns(uint) {
+        return calcDebt(fees[fee].chi, pie[loan]);
+    }
+    
+    function incDebt(uint fee, uint wad) private {
```
manuelpolzhofer:
```suggestion
    function incTotalDebt(uint fee, uint wad) private {
```

src/beans.sol
```diff
@@ -0,0 +1,176 @@
...
+    function deny(address usr) public auth note { wards[usr] = 0; }
+    modifier auth { require(wards[msg.sender] == 1); _; }
+    
+    // --- Data ---
+    // https://github.com/makerdao/dsr/blob/master/src/dsr.sol
+    struct Fee {
+        uint debt;
+        uint chi;
+        uint speed; // Accumulation per second
+        uint48 rho; // Last time the rate was accumulated
```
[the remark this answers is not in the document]
manuelpolzhofer: Just some ideas for better naming

src/beans.sol
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    mapping (uint => Fee) public fees;
+    mapping (uint => uint) public pie;
+
+    uint public debt;
```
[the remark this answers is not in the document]
manuelpolzhofer: I think we should only call the loan `debt`  => debt and refer fee and global debt as `totalDebt`.

Otherwise `incDebt` method names are confusing because we don't know which debt we increase

src/pile.sol
```diff
@@ -226,34 +123,29 @@ contract Pile is DSNote, TitleOwned {
...
+        beans.drip(loans_[loan].fee);
+    }
+
     // repay() a certain amount of token from the user to the Pile
     function repay(uint loan, uint wad) public owner(loan) note {
         // moves currency from usr to pile and reduces debt
         require(loans_[loan].balance == 0,"before repay loan needs to be withdrawn");
-        collect(loan);
+        uint fee = loans_[loan].fee;
+        beans.drip(fee);
```
manuelpolzhofer: why not use `collect(loan)` here?  it does exactly the same as the two lines.

src/pile.sol
```diff
@@ -226,34 +123,29 @@ contract Pile is DSNote, TitleOwned {
...
 
         tkn.transferFrom(msg.sender, address(this), wad);
-
-        uint chi = getChi(loan);
-        uint pie_ = rdiv(wad, chi);
-        loans_[loan].pie = sub(loans_[loan].pie, pie_);
-
-        decDebt(loans_[loan].fee, wad);
+        beans.decLoanDebt(loan, fee, wad);
         tkn.approve(lender, wad);
```
manuelpolzhofer: I think we can remove that line for Naxos. The pile should allow the `trancheManager` full control over the currency in pile.
I am not sure about the best pattern yet

src/beans.sol
```diff
@@ -0,0 +1,176 @@
...
+
+    function initLoanDebt(uint loan, uint fee, uint wad) public auth note {
+        drip(fee);
+        uint pie_ = calcPie(fees[fee].chi, wad);
+        pie[loan] = pie_;
+        incDebt(fee, wad);
+    }
+
+    function decLoanDebt(uint loan, uint fee, uint wad) public auth note {
+        uint pie_ = calcPie(fees[fee].chi, wad);
```
manuelpolzhofer: same here.

We also need a consistent use of `_`.

In the context of pile we used it for `delta chi's`.

src/pile.sol
manuelpolzhofer: let's discuss this face to face

src/beans.sol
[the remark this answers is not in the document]
ilinzweilin: left as rho, as we agreed not to rename existing vars

src/beans.sol
ilinzweilin: same here: left as speed, as we agreed not to rename existing vars

src/beans.sol
```diff
@@ -0,0 +1,175 @@
...
+    function rdiv(uint x, uint y) internal pure returns (uint z) {
+        z = add(mul(x, ONE), y / 2) / y;
+    }
+
+    function div(uint x, uint y) internal pure returns (uint z) {
+        z = x / y;
+    }
+
+    function incLoanDebt(uint loan, uint fee, uint wad) public auth note {
+        require(now >= fees[fee].rho);
```
manuelpolzhofer:
```suggestion
        require(now == fees[fee].rho);
```
Otherwise it is not required to call drip before.

src/beans.sol
```diff
@@ -0,0 +1,175 @@
...
+        pie[loan] = add(pie[loan], calcPie(fees[fee].chi, wad));
+        incTotalDebt(fee, wad);
+    }
+
+    function decLoanDebt(uint loan, uint fee, uint wad) public auth note {
+        require(now >= fees[fee].rho);
```
manuelpolzhofer:
```suggestion
        require(now == fees[fee].rho);
```

src/beans.sol
```diff
@@ -0,0 +1,175 @@
...
+        // compounding in seconds
+        uint latest = rmul(rpow(speed, now - rho, ONE), chi);
+        uint chi_ = rdiv(latest, chi);
+        uint wad = rmul(debt_, chi_) - debt_;
+        return (latest, chi_, wad);
+    }
+
+    // --- Fee Accumulation ---
+    function drip(uint fee) public {
+        if (now >= fees[fee].rho) {
```
manuelpolzhofer: I think this one should be actually
```suggestion
        if (now > fees[fee].rho) {
```

src/beans.sol
```diff
@@ -0,0 +1,175 @@
...
+            (uint latest, , uint wad) = compounding(fee);
+            fees[fee].chi = latest;
+            fees[fee].rho = uint48(now);
+            incTotalDebt(fee, wad);   
+        }
+    }
+
+    function burden(uint loan, uint fee) public view returns (uint) {
+        uint chi = fees[fee].chi;
+        if (now >= fees[fee].rho) {
```
manuelpolzhofer: this one is okay

src/beans.sol
```diff
@@ -0,0 +1,175 @@
...
+
+    function decLoanDebt(uint loan, uint fee, uint wad) public auth note {
+        require(now >= fees[fee].rho);
+        pie[loan] = sub(pie[loan], calcPie(fees[fee].chi, wad));
+        decTotalDebt(fee, wad);
+    }
+
+    function compounding(uint fee) public view returns (uint, uint, uint) {
+        uint48 rho = fees[fee].rho;
+        require(now >= rho);
```
manuelpolzhofer: this one is okay. because it is `view`

src/beans.sol
```diff
@@ -0,0 +1,175 @@
...
+        uint debt;
+        uint chi;
+        uint speed; // Accumulation per second
+        uint48 rho; // Last time the rate was accumulated
+    }
+
+    mapping (uint => Fee) public fees;
+    mapping (uint => uint) public pie;
+
+    uint public totalDebt;
```
charleenfei: wdyt about outstandingDebt? since this is what we are calling it internally?

src/test/beans.t.sol
```diff
@@ -0,0 +1,253 @@
...
+        beans.incLoanDebt(loan, fee, principal);
+
+        // 150 days later
+        hevm.warp(now + 1050 days); // produces max ~ chi 10^49
+        // debt ~ 10^27 * 10^49 =  10^76 (max uint is 10^77)
+        beans.drip(fee);
+    }
+
+    function checkDebt(uint loan, uint fee, uint should) public {
+        uint debt = beans.debtOf(loan, fee);
```
charleenfei: one edge case we could test for is updating of fee in the middle of a loan term. maybe we want to consider disallowing it?

src/beans.sol
ilinzweilin: I think total makes more sense to better distinguish from the loan or fee group debt. They are all outstanding

src/test/beans.t.sol
manuelpolzhofer: It is actually a feature. 

`fee`could for example be a risk group like `low` risk and `high` risk and the interest rates are changing.

src/beans.sol
ilinzweilin: @manuelpolzhofer true

PR #108, opened 2019-11-26
chore: moved actions library to tinlake core repo

charleenfei:
after issues with git submodules/dapp remapping errors, we decided to move the actions library to tinlake core.


manuelpolzhofer:
https://thoughts.t37.net/merging-2-different-git-repositories-without-losing-your-history-de7a06bba804

PR #111, opened 2019-12-05
feat: tag contract with unit tests

manuelpolzhofer:
I will split the collect contracts into multiple PR's


src/core/collect/tag.sol
```diff
@@ -0,0 +1,70 @@
...
+        wards[msg.sender] = 1;
+        pile = PileLike(pile_);
+        globalDiscount = ONE;
+    }
+
+    function price(uint loan) public returns (uint) {
+        pile.collect(loan);
+        if (discount[loan] == 0) {
+            return rmul(pile.debtOf(loan), globalDiscount);
+
```
charleenfei: space

src/core/test/mock/pile.sol
```diff
@@ -34,6 +35,10 @@ contract PileMock {
         return totalBalanceReturn;
     }
 
+    function debtOf(uint loan) public returns (uint) {
+        return debtOfReturn;
```
charleenfei: why not just return loan.debt?

src/core/test/mock/pile.sol
```diff
@@ -18,6 +18,7 @@ contract PileMock {
 
     // returns
     uint public debtReturn; function setLoanDebtReturn(uint debtReturn_) public {debtReturn=debtReturn_;}
+    uint public debtOfReturn; function setDebtOfReturn(uint debtOfReturn_) public {debtOfReturn=debtOfReturn_;}
```
charleenfei: why not use setLoanReturn? or if we want to be more modular, completely break up setLoanReturn into the three individual params on the loan.

manuelpolzhofer: because we currently using the following pattern:
the original contract as a method like
`function X(...) returns(uint)`

the mock contract has a method called 
`function setXReturn(uint)`

src/core/test/mock/pile.sol
manuelpolzhofer: because loan.debt is outdated and not existing anymore.

We only store pie.
I think somebody already adjusted the method in an unmerged PR

charleenfei: I think we should update this mock `Pile` contract to reflect the new `Loan` struct

src/core/collect/tag.sol
```diff
@@ -0,0 +1,69 @@
...
+    constructor(address pile_) public {
+        wards[msg.sender] = 1;
+        pile = PileLike(pile_);
+        globalDiscount = ONE;
+    }
+
+    function price(uint loan) public returns (uint) {
+        pile.collect(loan);
+        if (discount[loan] == 0) {
```
charleenfei: in what scenario would this check be used?

PR #112, opened 2019-12-09
test: use new mock Desk (Tranche Manager) in system test, update Pile mock, remove lender/backer, collateral and valve tests

### LINKED ISSUE #97: integration tests for tinlake actions
migrate/duplicate basic tests for borrow repay with CVT logic
- [x] happy path:  borrow / repay loan 
- [x] functional test repo can reuse the deployer contract

create on test happy path (borrow, repay)
- [x] using tinlake actions, tinlake-core borrower
- [x] new lender mock contract
- [ ] we might need our own deployer for it or we can pass an DeskFab for our lender mock

charleenfei:
Closes #97


src/core/test/system/system.t.sol
```diff
@@ -258,8 +230,8 @@ contract SystemTest is DSTest {
         return extra;
     }
 
-    function currLenderBal() public returns(uint) {
-        return tkn.balanceOf(lenderTokenAddr(address(deployer.lender())));
+    function currDeskBal() public returns(uint) {
```
manuelpolzhofer: I think we need to refactor it later because the desk won't hold currency. 
It will be in different tranches but I think it fine for now.

charleenfei: I will leave it like this for now but make a note to refactor when we have the rest of the lender side contracts in.

src/core/test/shelf.t.sol
```diff
@@ -116,11 +115,11 @@ contract ShelfTest is DSTest {
...
         nft.reset();
-        pile.setLoanReturn(0, 0, 0, 0);
+        pile.setLoanReturn(0, 0, 0);
         shelf.release(1, address(this));
-        assertEq(nft.from(), address(shelf));
-        assertEq(nft.to(), address(this));
-        assertEq(nft.transferFromCalls(), 1);
+//        assertEq(nft.from(), address(shelf));
+//        assertEq(nft.to(), address(this));
+//        assertEq(nft.transferFromCalls(), 1);
```
ilinzweilin: Is there a reason why this assertions are commented out?

charleenfei: no :)

PR #113, opened 2019-12-09
feat: collector & spotter

manuelpolzhofer:
please review #111 first


src/core/pile.sol
```diff
@@ -128,11 +128,20 @@ contract Pile is DSNote, TitleOwned {
         beans.drip(loans_[loan].fee);
```
charleenfei: I think we can think about taking this function abstraction out if it just calls only `beans.drip`

src/core/pile.sol
```diff
@@ -128,11 +128,20 @@ contract Pile is DSNote, TitleOwned {
...
-        require(loans_[loan].balance == 0,"before repay loan needs to be withdrawn");
+
+    // recovery used for defaulted loans_
+    function recovery(uint loan, uint wad) public auth {
+        doRepay(loan, wad);
+
+        uint fee = loans_[loan].fee;
+        uint loss = beans.debtOf(loan, fee);
+
+        beans.decLoanDebt(loan, loans_[loan].fee, loss);
```
charleenfei: why use `loans_[loan].fee` here if you already declared the variable above? or maybe we don't need to declare the variable.

src/core/pile.sol
```diff
@@ -146,6 +155,14 @@ contract Pile is DSNote, TitleOwned {
         tkn.approve(lender, wad);
     }
 
+    // repay() a certain amount of token from the user to the Pile
+    function repay(uint loan, uint wad) public owner(loan) note {
+        // moves currency from usr to pile and reduces debt
+        require(loans_[loan].balance == 0,"before repay loan needs to be withdrawn");
+        doRepay(loan, wad);
+
```
charleenfei: space

src/core/test/pile.t.sol
```diff
@@ -141,4 +141,5 @@ contract PileTest is DSTest {
         uint debt = pile.burden(loan);
         repay(loan, debt);
     }
+
```
charleenfei: space

src/core/collect/collector.sol
```diff
@@ -0,0 +1,77 @@
...
+    function file(bytes32 what, address addr) public auth {
+        if (what == "spotter") spotter = SpotterLike(addr);
+        else if (what == "tag") tag = TagLike(addr);
+        else if (what == "desk") desk = DeskLike(desk);
+        else if (what == "pile") pile = PileLike(pile);
+        else revert();
+    }
+
+    function collect(uint loan, address usr) public auth {
+        if(spotter.collectable(loan)){
```
charleenfei: !collectable

src/core/collect/spotter.sol
```diff
@@ -0,0 +1,117 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+pragma solidity >=0.4.24;
+
+contract ShelfLike {
+    function adjust(uint loan) public;
+    function shelf(uint loan) public returns (address, uint256, uint, uint);
```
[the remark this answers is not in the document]
manuelpolzhofer: I think we should generate a refactoring ticket for it.

src/core/pile.sol
manuelpolzhofer: Currently, it is easier because bean doesn't know the `fee` of a loan.

You would need to ask pile about the fee first

src/core/collect/collector.sol
```diff
@@ -0,0 +1,77 @@
...
+    function depend(bytes32 what, address addr) public auth {
+        if (what == "spotter") spotter = SpotterLike(addr);
+        else if (what == "tag") tag = TagLike(addr);
+        else if (what == "desk") desk = DeskLike(desk);
+        else if (what == "pile") pile = PileLike(pile);
+        else revert();
+    }
+
+    function collect(uint loan, address usr) public auth {
+        if(spotter.collectable(loan) == false){
```
charleenfei: !spotter.collectable(loan)

src/core/collect/test/integration.t.sol
```diff
@@ -0,0 +1,125 @@
...
+        assertEq(pile.loan(), loan);
+    }
+
+    function testCollectWithSpotter() public {
+        uint loan = 1; uint tokenId = 123;
+        uint price = 150 ether; uint debt = 100 ether;
+        setUpLoan(loan,tokenId, price, debt);
+
+        // 150% is enough threshold is 120%
+        assertTrue(spotter.seizable(loan) == false);
```
charleenfei: !spotter.seizable(loan)

src/core/collect/test/integration.t.sol
```diff
@@ -0,0 +1,125 @@
...
+
+        // increase debt to 115%
+        pile.setDebtOfReturn(130 ether);
+        assertTrue(spotter.seizable(loan) == true);
```
charleenfei: spotter.seizable(loan)

src/core/pile.sol
```diff
@@ -128,11 +128,18 @@ contract Pile is DSNote, TitleOwned {
...
-    // repay() a certain amount of token from the user to the Pile
-    function repay(uint loan, uint wad) public owner(loan) note {
-        // moves currency from usr to pile and reduces debt
-        require(loans_[loan].balance == 0,"before repay loan needs to be withdrawn");
+
+    // recovery used for defaulted loans_
+    function recovery(uint loan, uint wad) public auth {
+        doRepay(loan, wad);
+
+        uint loss = beans.debtOf(loan, loans_[loan].fee);
```
charleenfei: this is not actually the loss but the amount we have made up from recovering the defaulted loan. the loss would be the remaining debt after the below `beans.decLoanDebt` method is called.

ilinzweilin: @charly doRepay is calling the beans.decLoanDebt(). From my understanding... Manuel is just using the computed loss to decrease the the debt amount in beans... because no payment for this loan is expected anymore

src/core/pile.sol
```diff
@@ -128,11 +128,18 @@ contract Pile is DSNote, TitleOwned {
...
+        beans.decLoanDebt(loan, loans_[loan].fee, loss);
+    }
+
+    function doRepay(uint loan, uint wad) internal {
         collect(loan);
+
         uint fee = loans_[loan].fee;
         uint debt = beans.debtOf(loan, fee);
```
ilinzweilin: @manuelpolzhofer I think this check below prevents partial repay and consequently also recovery (which might be a partial repay)

src/core/collect/spotter.sol
```diff
@@ -0,0 +1,117 @@
...
+    function free(uint loan, address usr) public;
+}
+
+contract SPileLike {
```
ilinzweilin: why SPileLike and not PileLike

src/core/collect/spotter.sol
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    function nowPrice(uint loan) internal returns(uint) {
+        shelf.adjust(loan);
+        (,,uint price,) = shelf.shelf(loan);
+        return price;
+    }
+
+    function nowDebt(uint loan) internal returns(uint) {
+        pile.collect(loan);
```
ilinzweilin: could you also call: return pile.burden(loan) here?

manuelpolzhofer: I think we can implement it together with the cabinet contracts because it will change

src/core/pile.sol
manuelpolzhofer: yes. I just did it in that way to re-use the existing `decLoanDebt` method to set the loan to `zero`

Issue #94, opened 2019-11-20
add integration tests for collect loans


manuelpolzhofer:
using the lender mock
tests:  
happy path: borrow / repay loan  
default loan by debt to high  
default loan by updated appraisal (value loss)  
repay default loan should fail:

PR #120, opened 2019-12-12
test: split functional tests

manuelpolzhofer:
- functional tests have a base contract which can be used to create proxies and deploy the core contracts
- better naming for system tests


manuelpolzhofer:
PR is part of #94

PR #127, opened 2019-12-17
feat: desk

charleenfei:


src/core/desk.sol
```diff
@@ -0,0 +1,132 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+pragma solidity >=0.4.24;
+
+import "../../lib/dss-add-ilk-spell/lib/dss-deploy/lib/dss/src/lib.sol";
```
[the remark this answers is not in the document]
charleenfei: wip import, we don't need it

src/core/desk.sol
```diff
@@ -0,0 +1,132 @@
...
+            }
+        } else {
+            int wad = pile.want();
+            if (wad > 0) {
+                // transfer from reserve only how much the pile wants
+                t.operator.borrow(address(pile), uint(wad));
+            } else {
+                // this should take the money from the pile, repay and close the vault
+                distributor.makerRepay();
+                t.operator.repay(address(pile), uint(wad*-1));
```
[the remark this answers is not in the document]
charleenfei: yea, that is why i have makerRepay above this line. it was wip

src/core/desk.sol
```diff
@@ -0,0 +1,132 @@
...
+        // if capital should flow through, all funds in reserve should be moved in pile
+        if (flowThrough) {
+            if (t.operator.supplyActive()) {
+                // calculates how much money is in the reserve, transfers all of this balance to the pile
+                uint wadR = t.operator.balance();
+                t.operator.borrow(address(pile), uint(wadR));
+            }
+            if (t.operator.redeemActive()) {
+                // payout
+                distributor.waterfallRepay();
```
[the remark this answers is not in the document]
charleenfei: yep, this is what i mentioned in the standup that i wanted to do

src/core/desk.sol
```diff
@@ -0,0 +1,132 @@
...
+
+    function file(uint i, bytes32 what, bool data) public auth {
+        tranches[i].operator.file(what, data);
+    }
+
+    // --- Calls ---
+
+    function addTranche(uint ratio, address operator_) public auth {
+        Tranche memory t;
+        t.ratio = ratio;
```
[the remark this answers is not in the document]
charleenfei: yes, we can talk about what system makes sense (i was thinking about keeping it `uint` ie: 30) and then we can add a method for calculating the equity ratio on the Desk level. this allows us to extend for >2 or tranches inside of tranches. wdyt?

src/core/desk.sol
```diff
@@ -0,0 +1,132 @@
...
+    function balance(uint i) public auth {
+
+        // balance methods should:
+        // iterate through the tranches
+        // quant debt =  how much is
+
+        Tranche memory t = tranches[i];
+        // if capital should flow through, all funds in reserve should be moved in pile
+        if (flowThrough) {
+            if (t.operator.supplyActive()) {
```
ilinzweilin: supplyActive just controls if you can invest/put money in the tranche reserve, not if you can borrow from it. supply can be deactivated but money still can be borrowed from the reserve

src/core/desk.sol
```diff
@@ -0,0 +1,132 @@
...
+                // calculates how much money is in the reserve, transfers all of this balance to the pile
+                uint wadR = t.operator.balance();
+                t.operator.borrow(address(pile), uint(wadR));
+            }
+            if (t.operator.redeemActive()) {
```
ilinzweilin: same here: redeemActive just controls if investors can redeem their money from the tranche reserve, not if borrowers can repay

src/core/desk.sol
```diff
@@ -0,0 +1,132 @@
...
+                // payout
+                distributor.waterfallRepay();
+            }
+        } else {
+            int wad = pile.want();
+            if (wad > 0) {
+                // transfer from reserve only how much the pile wants
+                t.operator.borrow(address(pile), uint(wad));
```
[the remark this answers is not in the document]
ilinzweilin: I would move everything in the lines 121-129 to the distributor. Should be the only contract aware of the tranche logic and know from what tranche to take the money first and which tranche to repay fist

src/core/desk.sol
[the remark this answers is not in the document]
ilinzweilin: would give the method a more generic name then makerRepay 🙈

src/core/tranche/distributor.sol
```diff
@@ -0,0 +1,170 @@
...
+        for (uint i = 0; i < tranches.length; i++) {
+            uint wadR = tranches[i].operator.balance();
+            // this can be cast into uint bc wad should always be positive at this point
+            if (uint(wad) < wadR) {
+                tranches[i].operator.borrow(address(pile), uint(wad));
+                wad = 0;
+            } else {
+                tranches[i].operator.borrow(address(pile), uint(wadR));
+                wad = int(uint(wad) - wadR);
+            }
```
manuelpolzhofer: Here away with less code and a return if we are done.
```javascript
   if (uint(wad) < wadR) {
      tranches[i].operator.borrow(address(pile), uint(wad));
      return;
   }
   tranches[i].operator.borrow(address(pile), uint(wadR));
   wad = sub(wad, wadR);
```

src/core/tranche/distributor.sol
```diff
@@ -0,0 +1,170 @@
...
+        // pile wants money
+        if (wad > 0) {
+            pileWants();
+        } else if (wad < 0) {
+            pileHas();
+        }
+    }
+
+    function pileWants() private {
+        int wad = pile.want();
```
manuelpolzhofer: maybe just cast it once in the beginning?
```suggestion
        uint wad = uint(pile.want());
```

src/core/tranche/distributor.sol
```diff
@@ -0,0 +1,170 @@
...
+                wad = int(uint(wad) - wadR);
+            }
+        }
+    }
+
+    function pileHas() private {
+        int wad = pile.want();
+        // pile has extra money, this can be repaid into the reserve
+        // senior tranche should always be the last tranche in the array and be repaid first
+        for (int i = int(tranches.length - 1); i > int(-1); i--) {
```
manuelpolzhofer: That's the common way to iterate over an array starting from the last element.
```suggestion
        for (uint i = tranches.length - 1; i >= 0 ; i--) {
```

src/core/tranche/distributor.sol
```diff
@@ -0,0 +1,170 @@
...
+            QuantLike quant = tranches[uint(i)].operator.quant();
```
manuelpolzhofer:
```suggestion
            QuantLike quant = tranches[i].operator.quant();
```

src/core/tranche/distributor.sol
```diff
@@ -0,0 +1,170 @@
...
+            int wadD = int(quant.debt());
```
manuelpolzhofer: I think the debt is always positive or 0

src/core/tranche/distributor.sol
```diff
@@ -0,0 +1,170 @@
...
+            // positive number here means there is some debt in the senior tranche
+            if (wad < int(wadD*-1) && wadD > 0) {
+                // wadD cast to uint bc it should be positive here
+                tranches[uint(i)].operator.repay(address(pile), uint(wadD));
+                wad = wad + wadD;
+            } else {
+                // wad cast to uint bc it should be positive here
+                tranches[uint(i)].operator.repay(address(pile), uint(wad));
+                wad = 0;
+            }
```
manuelpolzhofer: I think the code can be simplified to:
```javascript
if (wadD >= wad) {
 tranches[uint(i)].operator.repay(address(pile), uint(wad));
 return;
}
 tranches[uint(i)].operator.repay(address(pile), uint(wadD));
 wad = sub(wad, wadD);
```

src/core/tranche/distributor.sol
```diff
@@ -0,0 +1,198 @@
...
+            // this should take the money from the pile, repay and close the vault
+            waterfallRepay();
+        }
+    }
+
+    function poolOpen() private {
+        int wad = pile.want();
+        // pile wants money
+        if (wad > 0) {
+            pileWants();
```
manuelpolzhofer: I would make `wad` here already positive and pass it as a parameter

src/core/tranche/distributor.sol
```diff
@@ -0,0 +1,198 @@
...
+        } else if (wad < 0) {
+            pileHas();
+        }
+    }
+
+    function pileWants() private {
+        int wad = pile.want();
+        uint max = calcMaxTake();
```
manuelpolzhofer: If it is generic max take shouldn't be `maxTake` called in the loop for every tranche?

src/core/tranche/distributor.sol
```diff
@@ -0,0 +1,198 @@
...
+
+    QuantLike public quant;
+}
+
+contract Distributor is DSNote {
+
+    // --- Tranches ---
+
+    struct Tranche {
+        bool senior;
```
manuelpolzhofer: If we are doing that we can't make it generic. 

We should either make it generic or only work for two tranches

src/core/tranche/distributor.sol
charleenfei: taking this suggestion into account but flipping the negatives/positives bc wad here is a negative number (pileHas)

manuelpolzhofer:
@ilinzweilin. I agree.  For the check, if we are doing a borrow or repay it depends on the case if we look at `pile.want()` so maybe this should be also in the distributor.

src/core/tranche/distributor.sol
```diff
@@ -0,0 +1,107 @@
...
+                // calculates how much money is in the reserve, transfers all of this balance to the pile
+                uint wadR = tranches[i].operator.balance();
+                tranches[i].operator.borrow(address(pile), uint(wadR));
+            }
+        }
+    }
+
+    // Takes all the money from the pile, pay sr tranche debt first, then pay jr tranche debt
+    function pileHas() private {
+        int wad = pile.want();
```
manuelpolzhofer: I would make it positive than we only need the casting once
```suggestion
        uint availableCurrency = uint(pile.want()*-1);
```

src/core/tranche/distributor.sol
```diff
@@ -0,0 +1,107 @@
...
+        for (uint i = tranches.length - 1; i >= 0; i--) {
+            QuantLike quant = tranches[i].operator.quant();
+            // should be positive number here, means there is some debt in the senior tranche, or 0
+            uint wadD = quant.debt();
```
manuelpolzhofer:
```suggestion
            uint trancheDebt = quant.debt();
```

src/core/tranche/distributor.sol
```diff
@@ -0,0 +1,107 @@
...
+            if (wadD >= uint(wad*-1)) {
```
manuelpolzhofer:
```suggestion
            if (trancheDebt >= availableCurrency) {
```

src/core/tranche/distributor.sol
```diff
@@ -0,0 +1,107 @@
...
+                tranches[i].operator.repay(address(pile), uint(wad*-1));
```
manuelpolzhofer:
```suggestion
                tranches[i].operator.repay(address(pile), availableCurrency);
```

src/core/tranche/distributor.sol
```diff
@@ -0,0 +1,107 @@
...
+                return;
+            }
+            tranches[i].operator.repay(address(pile), uint(wadD));
+            wad = int(wad) + int(wadD);
```
manuelpolzhofer:
```suggestion
            availableCurrency = availableCurrency - trancheDebt;
```

src/core/tranche/distributor.sol
```diff
@@ -0,0 +1,107 @@
...
+        int wad = pile.want();
+        for (uint i = tranches.length - 1; i >= 0; i--) {
+            QuantLike quant = tranches[i].operator.quant();
+            // should be positive number here, means there is some debt in the senior tranche, or 0
+            uint wadD = quant.debt();
+            if (wadD >= uint(wad*-1)) {
+                tranches[i].operator.repay(address(pile), uint(wad*-1));
+                return;
+            }
+            tranches[i].operator.repay(address(pile), uint(wadD));
```
manuelpolzhofer:
```suggestion
            tranches[i].operator.repay(address(pile), trancheDebt);
```

src/core/tranche/distributor.sol
```diff
@@ -0,0 +1,107 @@
...
+        t.operator = OperatorLike(operator_);
+        tranches.push(t);
+    }
+
+    function ratioOf(uint i) public auth returns (uint) {
+        return tranches[i].ratio;
+    }
+
+    // if capital should flow through, all funds in reserve should be moved in pile
+    function handleFlow(bool flowThrough, bool poolClosing) public auth {
```
manuelpolzhofer: I think we don't need to pass the flags. We will use an own contract with inheritance for the flags similar to lightswitch

src/core/tranche/orchestrator.sol
```diff
@@ -0,0 +1,42 @@
...
+
+import "ds-note/note.sol";
+
+import {Diffuser} from "./diffuser.sol";
+import {Distributor} from "./distributor.sol";
+
+contract Orchestrator is DSNote{
+
+    Diffuser public diffuser;
+    Distributor public distributor;
```
manuelpolzhofer: If we are implementing the Orchestrator. 
@ilinzweilin  might be right and we never need both in one deployment.

I would use here a `DistributorLike` abstract contract/interface

src/core/tranche/orchestrator.sol
```diff
@@ -0,0 +1,42 @@
...
+
+    // --- Auth ---
+    mapping (address => uint) public wards;
+    function rely(address usr) public auth note { wards[usr] = 1; }
+    function deny(address usr) public auth note { wards[usr] = 0; }
+    modifier auth { require(wards[msg.sender] == 1); _; }
+
+    function handleFlow(bool flowThrough, bool poolClosing) public auth {
```
manuelpolzhofer: I think we don't need the `auth` here.

Only if we add `depends` method for setting the two distributors

src/core/tranche/distributor/maker.sol
```diff
@@ -0,0 +1,92 @@
...
+            repayTranches();
+        }
+    }
+
+    function borrowFromTranches(uint wad) private {
+        uint borrowAmount = wad;
+        for (uint i = 0; i < manager.trancheCount(); i++) {
+            OperatorLike o = OperatorLike(manager.operatorOf(i));
+            uint trancheReserve = o.balance();
+            if (borrowAmount < trancheReserve) {
```
manuelpolzhofer:
```suggestion
            if (borrowAmount <= trancheReserve) {
```
I think needs to be like that. Otherwise considering the edge case:

`borrowAmount = 100, trancheReserve = 100`

The if-condition would be false, and we would borrow 100 from the first tranche.
The loop would continue and we would ask the next tranche to borrow `0`

src/core/tranche/distributor/maker.sol
```diff
@@ -0,0 +1,92 @@
...
+    // in order to maintain the equity ratio which has been set by the pool manager.
+
+    // max_take =  (Equity.Reserve + Equity.Debt)/Equity.Ratio * Senior.Ratio - Senior.Debt
+
+    function calcMaxTake() public auth returns (uint) {
+        OperatorLike oE = OperatorLike(manager.operatorOf(0));
+        OperatorLike oS = OperatorLike(manager.operatorOf(1));
+
+        uint wadER = oE.balance();
+        uint wadED =  oE.debt();
```
manuelpolzhofer: I think this one is wrong. We need the `real` debt not the expected `debt` from equity tranche.
`oE.debt()` is not covering the losses.

src/core/tranche/distributor/maker.sol
```diff
@@ -0,0 +1,92 @@
...
+            borrowAmount = sub(borrowAmount, trancheReserve);
+        }
+    }
+
+    // Note: assumes two tranche setup for now
+
+    // max_take is how much liquidity can be taken out from a specific tranche, given the current equity reserve/equity debt,
+    // in order to maintain the equity ratio which has been set by the pool manager.
+
+    // max_take =  (Equity.Reserve + Equity.Debt)/Equity.Ratio * Senior.Ratio - Senior.Debt
```
manuelpolzhofer: If we are using this formula we are considering the `Senior.Reserve` always to be 0.

We just need to write it down somewhere

src/core/tranche/distributor/maker.sol
```diff
@@ -0,0 +1,92 @@
...
+            if (borrowAmount < trancheReserve) {
+                o.borrow(address(manager.pile), borrowAmount);
+                return;
+            }
+            o.borrow(address(manager.pile), trancheReserve);
+            borrowAmount = sub(borrowAmount, trancheReserve);
+        }
+    }
+
+    // Note: assumes two tranche setup for now
```
manuelpolzhofer: We should consider a common pattern for method comments. (if we need additional text to explain)

src/core/tranche/manager.sol
```diff
@@ -0,0 +1,104 @@
...
+
+    function file(bytes32 what, bool data) public auth {
+        if (what == "flowThrough") { flowThrough = data; }
+        if (what == "poolClosing") { poolClosing = data; }
+    }
+
+    // --- Calls ---
+
+    // TIN tranche should always be added first
+    function addTranche(uint ratio, address operator_) public auth {
```
manuelpolzhofer: We need a comment for the ratio.
What are we using here, RAD?  For a ratio of 70% is it `7 * 10^26` (0.7) or `70 * 10^27` (70)
```suggestion
    function addTranche(uint ratio, address operator_) public auth {
```

src/core/tranche/distributor/maker.sol
```diff
@@ -0,0 +1,92 @@
...
+            poolOpen();
+        }
+    }
+
+    function poolOpen() private {
+        int wad = manager.checkPile();
+        if (wad > 0) {
+            borrowFromTranches(uint(wad));
+        } else if (wad < 0) {
+            repayTranches();
```
manuelpolzhofer:
```suggestion
            repayTranches(uint(wad));
```

manuelpolzhofer: for consistency with borrowFromTranches

src/core/tranche/distributor/maker.sol
```diff
@@ -0,0 +1,92 @@
...
+        }
+    }
+
+    function borrowFromTranches(uint wad) private {
+        uint borrowAmount = wad;
```
manuelpolzhofer:
```suggestion

```

src/core/tranche/distributor/maker.sol
```diff
@@ -0,0 +1,92 @@
...
+    function poolOpen() private {
+        int wad = manager.checkPile();
+        if (wad > 0) {
+            borrowFromTranches(uint(wad));
+        } else if (wad < 0) {
+            repayTranches();
+        }
+    }
+
+    function borrowFromTranches(uint wad) private {
```
manuelpolzhofer:
```suggestion
    function borrowFromTranches(uint borrowAmount) private {
```

src/core/tranche/distributor/maker.sol
```diff
@@ -0,0 +1,92 @@
...
+import "../../lightswitch.sol";
+
+contract MakerDistributor is Distributor, Switchable {
+
+    function balance() public auth {
+
+//        require(flowThrough==false);
+
+        if (manager.poolClosing()) {
+            repayTranches();
```
manuelpolzhofer: I am a bit confused. Are we deciding if we repay here with a flag or with the return of pileCheck?
Currently, it is both.

src/core/tranche/distributor.sol
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    function depend (bytes32 what, address addr) public auth {
+        if (what == "manager") { manager = ManagerLike(addr); }
+        else revert();
+    }
+
+    function repayTranches() public auth {
+        uint availableCurrency = uint(manager.checkPile()*-1);
+        for (uint i = manager.trancheCount() - 1; i >= 0; i--) {
```
manuelpolzhofer: nice 👍

src/core/tranche/distributor/maker.sol
```diff
@@ -0,0 +1,92 @@
...
+            borrowFromTranches(uint(wad));
+        } else if (wad < 0) {
+            repayTranches();
+        }
+    }
+
+    function borrowFromTranches(uint wad) private {
+        uint borrowAmount = wad;
+        for (uint i = 0; i < manager.trancheCount(); i++) {
+            OperatorLike o = OperatorLike(manager.operatorOf(i));
```
manuelpolzhofer: good idea to call it `o` 👍 golang style

src/core/tranche/distributor/trad.sol
```diff
@@ -0,0 +1,37 @@
...
+    function balance() public auth {
+
+//        require(flowThrough);
+
+        if (manager.poolClosing()) {
+            repayTranches();
+        } else {
+            for (uint i = 0; i < manager.trancheCount(); i++) {
+                OperatorLike o = OperatorLike(manager.operatorOf(i));
+                uint availableCurrency = o.balance();
```
manuelpolzhofer: maybe with a check?
```javascript
if (availableCurrency > 0) {
o.borrow(address(manager.pile), uint(availableCurrency));
}
```

src/core/tranche/manager.sol
charleenfei: 70 * 70ˆ27 (RAD)

src/core/tranche/distributor/line.sol
```diff
@@ -0,0 +1,85 @@
...
+            if (borrowAmount <= trancheReserve) {
+                o.borrow(address(manager.pile), borrowAmount);
+                return;
+            }
+            o.borrow(address(manager.pile), trancheReserve);
+            borrowAmount = sub(borrowAmount, trancheReserve);
+        }
+    }
+
+    // Note: assumes two tranche setup for now
```
charleenfei: Note: max_take formula will be updated and moved with losses PR

src/core/tranche/manager.sol
```diff
@@ -0,0 +1,111 @@
...
+        tranches.push(t);
+    }
+
+    function balance() public auth {
+        if (poolClosing) {
+            distributor.repayTranches(uint(pile.want()*-1));
+        } else {
+            distributor.balance();
+        }
+    }
```
manuelpolzhofer: We have two ways how we can implement it.

TrancheManager has balance method and calls `repay` or `borrow` on the distributor and is called from here.

Which means we can remove the balance method from the two Distributors.

The problem is we would need already a different logic here depending on the tranche manager because `pile.want` is already specific to one distributor implementation.

The second option is to remove the balance method from here or only call balance on the distributor. 
The balance method in the distributor would check if it is repay or borrow.

I think the second option is better to keep the logic in the distributor.

src/core/tranche/distributor/line.sol
```diff
@@ -0,0 +1,85 @@
...
+        int wad = manager.checkPile();
+        if (wad > 0) {
+            borrowFromTranches(uint(wad));
+        } else if (wad < 0) {
+            // takes extra money from Pile and repays tranche debt to avoid accumulation of fees
+            repayTranches(uint(wad*-1));
+        }
+    }
+
+    function borrowFromTranches(uint borrowAmount) private {
```
manuelpolzhofer: I have an idea. I  think we can move this method to the Distributor base contract as well.

The two different distributors would only a different `balance` implementation. 

The LOC distributor would call borrow with the borrowAmount and the fixed distributor could call this method
with `borrowAmount = uint(-1)` which would result in borrowing all the available currency from all tranches.

What do you think?

manuelpolzhofer: We could say the LOC is our default implementation of the distributor with the balance method. For the fixed one we use inheritance and overwrite the balance method. That's it.

We could remove the LOC contract.

manuelpolzhofer: Even the balance method would be similar. For the repay we always use all the currency available in the pile.

Only for the borrow case the currencyAmount is different. In one distirbutor based on pile.Want in the other one `uint(-1)` to take it all.

We might could only have one Distibutor with a `takeItAll` flag for the two ways to get the currencyAmount

manuelpolzhofer:
@charleenfei  I don't have permission to push to your PR. I will close this one and reopen

PR #129, opened 2019-12-18
fix: pile recovery with usr and collector functional tests

manuelpolzhofer:


src/core/pile.sol
```diff
@@ -130,14 +130,14 @@ contract Pile is DSNote, TitleOwned {
 
 
     // recovery used for defaulted loans_
-    function recovery(uint loan, uint wad) public auth {
-        doRepay(loan, wad);
+    function recovery(uint loan, address usr, uint wad) public auth {
```
manuelpolzhofer: We need to pass the `usr` here because `msg.sender` would be the collector contract but we want to take the currency from the keeper

src/core/shelf.sol
```diff
@@ -105,8 +104,8 @@ contract Shelf is TitleOwned {
         adjust(loan);
     }
 
-    function free(uint loan, address usr) public auth {
-        move(loan, shelf[loan].registry, shelf[loan].tokenId, usr);
+    function free(uint loan, address usr) public auth  {
```
manuelpolzhofer: we can't use move here because move has the `loanOwner` modifier.

src/test/basic.t.sol
```diff
@@ -33,46 +33,15 @@ contract BasicTest is FunctionalTest {
...
-        mintCollateralNFT(proxy_, tokenId);
-        uint loan = systemTest.whitelist(tokenId, tinlake.collateralNFT_, principal, appraisal, proxy_, fee);
-        borrower.approve(proxy_,  actions_, tinlake.collateralNFT_, tinlake.shelf_, tokenId);
-        borrower.borrow(proxy_, actions_, tinlake.desk_, tinlake.pile_, tinlake.shelf_, loan, proxy_);
+        uint loan = whitelistAndBorrow(tokenId, principal, appraisal, fee);
 
         systemTest.hevm().warp(now + 10 days);
-
-        uint extra = 100000000000 ether;
-        // add liquidity for repayment
```
charleenfei: 👍

src/test/collect.t.sol
```diff
@@ -25,19 +26,45 @@ import { Actions } from "../actions/actions.sol";
...
+        (uint tokenId, uint principal, uint appraisal, uint fee) = systemTest.defaultLoan();
+        uint loan = whitelistAndBorrow(tokenId, principal, appraisal, fee);
+
+        CollectDeployer collectDeployer = CollectDeployer(address(systemTest.deployer().collectDeployer()));
+
+        // intial threshold and ratio at 120%
+        assertEq(collectDeployer.spotter().threshold(), rdiv(appraisal, principal));
+
+        bool seizable = collectDeployer.spotter().seizable(loan);
+        assertTrue(seizable==false);
```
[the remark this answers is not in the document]
manuelpolzhofer: i would prefer `seizable==false` it is easier to read.

PR #131, opened 2019-12-19
test: added functionals tests for collector

manuelpolzhofer:
please review #129 first


src/test/collect.t.sol
```diff
@@ -25,19 +26,76 @@ import { Actions } from "../actions/actions.sol";
...
+
+        // allow test to collect NFT's
+        systemTest.admin().doAddKeeper(address(this));
+
+        assertTrue(systemTest.deployer().pile().Debt() != 0);
+
+        setUpRepayLiquidity(address(this), 2000 ether);
+        // collect NFT
+        collectDeployer.collector().collect(loan, address(this));
+
```
ilinzweilin: maybe also check the currency balance of the keeper address(this)

src/test/collect.t.sol
```diff
@@ -25,19 +26,76 @@ import { Actions } from "../actions/actions.sol";
...
+        // intial threshold and ratio at 120%
+        assertEq(collectDeployer.spotter().threshold(), rdiv(appraisal, principal));
+
+        bool seizable = collectDeployer.spotter().seizable(loan);
+        assertTrue(seizable==false);
+        return (loan, tokenId, principal, appraisal, fee);
+
+    }
+
```
ilinzweilin: I would also add a test. trying to collect a healthy loan

manuelpolzhofer: good idea will add both tests

src/test/functional.t.sol
```diff
@@ -109,4 +113,40 @@ contract FunctionalTest is DSTest {
...
+        uint debt = PileLike(tinlake.pile_).debtOf(loan);
+        borrower.close(proxy_, actions_, tinlake.desk_, tinlake.pile_, tinlake.shelf_, loan, proxy_);
+
+        assertEq(ERC20Like(tinlake.currency_).balanceOf(proxy_), extra + principal -  debt);
+        assertEq(PileLike(tinlake.pile_).balanceOf(loan), 0);
+        assertEq(Title(tinlake.collateralNFT_).ownerOf(tokenId), proxy_);
+
+    }
+
+    function whitelistAndBorrow(uint tokenId, uint principal, uint appraisal, uint fee) public returns (uint) {
```
charleenfei: 👍

PR #132, opened 2019-12-19
feat: make it possible to include losses into redeem & supply

ilinzweilin:
- adjust formulas to compute payouts & slices for redeem() / supply() according to: 
https://centrifuge.hackmd.io/ODh184sJTk-2P50xpM39eQ?both
- compute tokenPrice not based on iSupply anymore, but based on the tranche asset valuation (provided by the tranche manager) and the total supply of tokens
- depending on the calculation tranche asset valuation can include losses


src/core/tranche/slicer.sol
```diff
@@ -17,69 +17,59 @@ pragma solidity >=0.4.24;
...
-    function updateSupplyRate(uint borrowSpeed, uint debt, uint reserve) public note auth {
-        require (borrowSpeed > 0);
-        if (now >= iSupply.rho) {
-            drip();
-        }
-        uint ratio = rdiv(debt, add(reserve, debt));
-        uint borrowSpeed_ = sub(borrowSpeed, ONE);
-        iSupply.speed = add(rmul(borrowSpeed_, ratio), ONE);
+    function calcTokenPrice(uint tokenSupply, uint totalAssets) internal pure returns (uint) {
+        return rdiv(totalAssets, tokenSupply);
```
manuelpolzhofer: `uint tokenSupply, uint totalAssets` are both `wad` 10^18 and the tokenPrice is dominated in `rad`.

I think we should signal that with comments

src/core/tranche/slicer.sol
```diff
@@ -17,69 +17,59 @@ pragma solidity >=0.4.24;
...
+        return calcTokenPrice(totalSupply, totalAssets);
     }
 
-    function drip() public note auth {
-         if (now >= iSupply.rho) {
-            iSupply.chi = rmul(rpow(iSupply.speed, now - iSupply.rho, ONE),  iSupply.chi );
-            iSupply.rho = uint48(now);
-        }
+    function calcPayout(uint tokenAmount, uint tokenPrice) internal pure returns (uint) {
+        return rmul(tokenAmount, tokenPrice);
```
manuelpolzhofer: because tokenPrice is in RAD we need `rmul` correct?

src/core/tranche/slicer.sol
```diff
@@ -17,69 +17,59 @@ pragma solidity >=0.4.24;
...
     }
 
+    function calcSlice(uint currencyAmount, uint tokenPrice) internal pure returns (uint) {
+        return rdiv(currencyAmount, tokenPrice);
```
manuelpolzhofer: same here. let's add some comments

src/core/tranche/operator.sol
```diff
@@ -78,9 +73,9 @@ contract Operator is DSNote {
 
     function supply(address usr, uint currencyAmount) public note auth {
         require (supplyActive);
-        uint tokenAmount = slicer.calcSlice(currencyAmount);
+        uint tokenAmount = slicer.getSlice(currencyAmount);
```
manuelpolzhofer: It is a bit confusing we have `getSlice` and `sliceOf` which sounds really similar but they are doing different things

src/core/tranche/quant.sol
```diff
@@ -17,73 +17,90 @@ pragma solidity >=0.4.24;
...
+
+    function updateBorrowRate() public note auth {
+        if (supplyRateFixed && supplyRate > 0) {
+            if (now >= borrowRate.rho) {
+                drip();
+            }
+            uint balance = reserve.balance();
+            uint supplyRate_ = sub(supplyRate, ONE);
+            uint ratio = rdiv(add(balance, debt), debt); 
+            borrowRate.speed = add(rmul(ratio, supplyRate_), ONE);
```
manuelpolzhofer: In general the formular is correct but don't we run into our decimal number issue here.
assuming supplyRate is something like `1000000564701133626865910626` => 10^27.
`1000000564701133626865910626` - ONE would result in `564701133626865910626`.

Didn't we come up already with a solution for that?

src/core/tranche/operator.sol
ilinzweilin: renamed sliceOf(address usr) to tokenBalance(address usr)

ilinzweilin:
since we are moving to descriptive names, I would pick: tranche, TrancheManager, rate

src/core/tranche/quant.sol
```diff
@@ -17,79 +17,96 @@ pragma solidity >=0.4.24;
...
-        iBorrow.speed = add(rmul(ratio, supplySpeed_), ONE);
     }
 
-    function drip() public note auth{
-        if (now >= iBorrow.rho) {
-            (uint latest, , uint wad) = compounding();
-            iBorrow.chi = latest;
-            iBorrow.rho = uint48(now);
-            debt = add(debt, wad);   
+    function setFixedSupplyRate(bool fixed_) public note auth {
```
charleenfei: this could be in the `file` method

src/core/tranche/quant.sol
```diff
@@ -17,79 +17,96 @@ pragma solidity >=0.4.24;
...
+        supplyRateFixed = fixed_;
+    }
+
+    function updateBorrowRate() public note auth {
+        if (supplyRateFixed && supplyRate > 0) {
+            if (now >= borrowRate.lastUpdated) {
```
charleenfei: this `if` statement is repeated in drip(), i think it's redundant

src/core/tranche/quant.sol
```diff
@@ -17,79 +17,96 @@ pragma solidity >=0.4.24;
...
+                drip();
+            }
+            uint balance = reserve.balance();
+            uint supplyRate_ = sub(supplyRate, ONE);
+            uint ratio = rdiv(add(balance, debt), debt);
```
charleenfei: doesn't have to be in this PR, but here is one example of an area that would be really great here to have a legend of which variables are `rad` or `wad`

src/core/tranche/quant.sol
```diff
@@ -17,79 +17,96 @@ pragma solidity >=0.4.24;
...
+            uint balance = reserve.balance();
+            uint supplyRate_ = sub(supplyRate, ONE);
+            uint ratio = rdiv(add(balance, debt), debt); 
+            borrowRate.speed = add(rmul(ratio, supplyRate_), ONE);
         }
     }
 
     function updateDebt(int wad) public note auth  {
-        if (now >= iBorrow.rho) {
+        if (now >= borrowRate.lastUpdated) {
```
charleenfei: this `if` statement is repeated in drip() and redundant

src/core/tranche/quant.sol
```diff
@@ -17,79 +17,96 @@ pragma solidity >=0.4.24;
...
             drip();
         }
         debt = uint(int(debt) + int(wad));
```
charleenfei: actually this whole function is contained in `drip`

src/core/tranche/quant.sol
```diff
@@ -17,79 +17,96 @@ pragma solidity >=0.4.24;
...
+            borrowRate.lastUpdated = uint48(now);
+            debt = add(debt, wad);   
+        }
+    }
+
     function compounding() internal view returns (uint, uint, uint) {
-        uint48 rho = iBorrow.rho;
-        require(now >= rho);
-        uint speed = iBorrow.speed;
+        uint48 lastUpdated = borrowRate.lastUpdated;
```
charleenfei: nitpick but i don't think this variable necessarily has to be declared

src/core/tranche/quant.sol
manuelpolzhofer: I don't think so. 

drip => increases current debt based on interest
updateDebt => increase/decrease debt (based on repays and borrows) + drip


but we can remove the if condition everywhere and keep one in the drip 👍

src/core/tranche/quant.sol
ilinzweilin: it requires a bool and the file method is generic for strings

src/core/tranche/quant.sol
charleenfei: 👍

src/core/tranche/quant.sol
ilinzweilin: I prefer to declare vars, if I am using them more then once to increase readability... no hard feelings, so let's agree on a convention here :)

src/core/tranche/quant.sol
ilinzweilin: yes, exactly. updateDebt is actually called when new money is borrowed

src/core/tranche/quant.sol
manuelpolzhofer: @ilinzweilin it is cheaper the other way :)

PR #138, opened 2019-12-26
fix: quant updateBorrowRate zero debt

manuelpolzhofer:
- added some tests for updaetBorrowRate edge cases
- fix: handle debt == 0 edge case

I will invest some more time on that method  in the next days


ilinzweilin:
cool!

PR #142, opened 2019-12-30
Collector Refactor

lucasvo:
I'm starting to simplify the collector a bit


lucasvo:
@manuelpolzhofer I've started simplifying the collector logic a bit. 

I've made a few changes to move logic into external modules. The Tag can be removed as it is right now fully controlled by a ward anyway. We can just have a simple file(uint loan, address usr, uint wad) method that allows another contract to determine who may get collect the NFT. If the user then calls `collect()` it will transfer `wad` token form the user's balance and move the NFT into the users account.

Always keeping the NFT in the shelf: If we move the NFT out of the shelf into the collector/spotter, we'd have to duplicate a lot of data: the token id, registry address and implement methods inside of the collector to deal with transfers, set approvals or do some other things. Instead I've opted to introduce a "lock" that the collector can set on the NFT to prevent the loan owner from withdrawing the NFT. The NFT thus can stay in the Shelf until the collector is ready to move it into the keepers account.

I've also changed how the Tag works: instead of having a separate contract that stores data and has logic, I've created a "Tag" in the Collector contract that simply keeps track of a value and user that is authorized to collect the token at that value. If one wanted to implement an auction, discounting with a price curve or other things, the Tag contract would simply have to be authorized on the Collector contract to set the Tag value.

src/core/shelf.sol
```diff
@@ -103,19 +105,21 @@ contract Shelf is TitleOwned {
...
     function deposit (uint loan, address usr) public owner(loan) {
         NFTLike(shelf[loan].registry).transferFrom(usr, address(this), shelf[loan].tokenId);
         pile.borrow(loan, shelf[loan].principal);
         shelf[loan].principal = 0;
         adjust(loan);
     }
 
+    // Used by the collector
+    function claim(uint loan, address usr) public auth {
+        NFTLike(shelf[loan].registry).transferFrom(address(this), usr, shelf[loan].tokenId);
```
manuelpolzhofer: Why did you remove the adjust?

We have two options either we call adjust after each interaction which changes the bags or we don't guarantee an updated bags value and each external call needs to call adjust before.

src/core/shelf.sol
```diff
@@ -52,12 +52,13 @@ contract Shelf is TitleOwned {
...
-        uint price; //
-        uint principal; //
-        uint initial; //
+        uint price;
+        uint principal;
+        uint initial;
     }
 
     mapping (uint => Loan) public shelf;
+    mapping (uint => uint) public locks;
```
manuelpolzhofer: Why not `bool` why not inside the loans struct?

src/core/collect/collector.sol
```diff
@@ -34,44 +26,62 @@ contract PileLike {
...
+        else if (what == "liquidation") liquidation = RegistryLike(addr);
         else revert();
     }
 
-    function collect(uint loan, address usr) public auth  {
-        if(spotter.collectable(loan) == false){
-            spotter.seizure(loan);
-        }
+    // --- Collector ---
+    function file(uint loan, address usr, uint wad) auth {
```
manuelpolzhofer:
```suggestion
    function file(uint loan, address usr, uint wad) public auth {
```

manuelpolzhofer: For a later auction a `usr` is required but how would it work if we don't care which of the whitelisted address would buy th nft?

The external contract needs to be called to pass my address here?

src/core/shelf.sol
```diff
@@ -94,7 +94,9 @@ contract Shelf is TitleOwned {
 
     // Move the NFT out of the shelf
     function move(uint loan, address registry_, uint nft_, address to) public owner(loan) {
+        require(locks[loan] == 0, "nft-locked");
         NFTLike(registry_).transferFrom(address(this), to, nft_);
+        // TODO: Shouldn't this only be allowed if debt == 0?
```
manuelpolzhofer: good catch 🥇 

I justed looked it up it happened during the refactoring from `auth`. The method should be internal.

src/core/shelf.sol
```diff
@@ -94,7 +94,9 @@ contract Shelf is TitleOwned {
 
     // Move the NFT out of the shelf
     function move(uint loan, address registry_, uint nft_, address to) public owner(loan) {
```
manuelpolzhofer:
```suggestion
    function move(uint loan, address registry_, uint nft_, address to) internal {
```

src/core/collect/collector.sol
```diff
@@ -14,17 +14,9 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-pragma solidity >=0.4.24;
```
manuelpolzhofer: We plan to change the version. Please in a separated PR with adjusted travis script

src/core/shelf.sol
lucasvo: For this reason: https://gist.github.com/lucasvo/5cf682c29325e450b12b4f963c856cf1

src/core/collect/collector.sol
lucasvo: I've updated travis to support 0.5.12, we can fix all the other versions in 0.5.12 later. I agree.

src/core/collect/collector.sol
lucasvo: we could add an edge case where if wad > 0 and usr == 0 then any usr could buy it?

src/core/shelf.sol
```diff
@@ -46,40 +42,63 @@ contract Shelf is TitleOwned {
...
+        nftlookup[nft] = loan;
+
+        shelf[loan].registry = registry;
+        shelf[loan].tokenId = token;
+
+        return loan;
     }
 
+    function close(uint loan) public owner(loan) {
+        require(pile.debtOf(loan) == 0, "outstanding-debt"); // TODO: only allow closing of a loan that isn't active anymore. maybe there is a better criteria
```
ilinzweilin: in case of async borrowing the loan might have a balance that has not been withdrawn yet. If we don't decrease it, we might end up taking more money from the lending side then required.

src/core/shelf.sol
```diff
@@ -92,42 +111,38 @@ contract Shelf is TitleOwned {
...
-    function release (uint loan, address usr) public owner(loan) {
-        uint debt = pile.debtOf(loan);
-        require(debt == 0, "debt");
-        move(loan, shelf[loan].registry, shelf[loan].tokenId, usr);
-        adjust(loan);
+    // --- Currency actions ---
+    function borrow(uint loan, uint wad) public owner(loan) {
+        debt.accrue(loan);
+        // principal.borrow(loan, wad); TODO: reentrancy
+        debt.inc(loan, wad);
```
ilinzweilin: we need to increase balance here as well for the balance() function of Desk/TrancheManager

src/core/debt_register.sol
```diff
@@ -107,79 +114,108 @@ contract DebtRegister is DSNote {
...
+        uint index_ = rmul(rpow(speed, now - rho, ONE), index);
+        uint debt = fromChi(chi_, index);
+        require(index != 0);
+        index_delta = rdiv(index_, index);
+        delta = rmul(chi_, index_) - debt;
 
-    function incTotalDebt(uint rate, uint wad) private {
-        rates[rate].debt = add(rates[rate].debt, wad);
-        totalDebt = add(totalDebt, wad);
+        return (index_, index_delta, delta);
```
manuelpolzhofer: why are you returning the `index_delta`. I think it is not needed?

src/core/debt_register.sol
```diff
@@ -107,79 +114,108 @@ contract DebtRegister is DSNote {
...
-        rates[rate].debt = sub(rates[rate].debt, wad);
-        totalDebt = sub(totalDebt, wad);
+    // drip updates the index of the rate category by compounding the interest and
+    // updates the total debt
+    function drip(uint rate) public {
+        if (now >= rates[rate].rho) {
+            (uint index, , uint delta) = compound(rate);
+            rates[rate].index = index;
+            rates[rate].rho = uint48(now);
+            total = add(total, delta);
```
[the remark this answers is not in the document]
lucasvo: @manuelpolzhofer @ilinzweilin sorry for the confusion. This is actually why I changed to storing pie and not debt in the rate. If we do so, we save ourselves updating `Rate.pie` because it doesn't change if you call drip(). My changes I'm about to push will make this clearer.

src/core/debt_register.sol
```diff
@@ -107,79 +114,108 @@ contract DebtRegister is DSNote {
...
+        drip(rate);
+        drip(rate_);
+        uint chi_ = chi[loan];
+        uint debt = fromChi(rates[rate].index, chi_);
+        rates[rate].chi = sub(rates[rate].chi, chi_);
+        chi[loan] = toChi(rates[rate_].index, debt);
+        rates[rate_].chi = add(rates[rate_].chi, chi[loan]);
+        // TODO decTotalDebt(currentRate, debt);
+        // TODO incTotalDebt(newRate, debt);
     }
```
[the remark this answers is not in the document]
lucasvo: Yup true. That method is still broken though. I will address that later.

src/core/shelf.sol
```diff
@@ -46,40 +42,63 @@ contract Shelf is TitleOwned {
...
+    function token(uint loan) public view returns (address registry, uint nft) {
+        return (shelf[loan].registry, shelf[loan].tokenId);
+    }
+
+    // --- Shelf: Creation and closing of a loan ---
+    function issue(address registry, uint token) public returns (uint) {
+        require(NFTLike(registry).ownerOf(token) == msg.sender, "nft-not-owned");
+
+        bytes32 nft = keccak256(abi.encodePacked(registry, token));
+        require(nftlookup[nft] == 0, "nft-in-use");
```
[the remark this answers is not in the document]
lucasvo: :+1:

src/core/shelf.sol
lucasvo: yeah, I would like to discuss the balance handling again, I think there are a few of these issues still in the code base.

src/core/collect/collector.sol
```diff
@@ -14,64 +14,85 @@
...
-        }
+    // --- Collector ---
+    function file(uint loan, address usr, uint wad) public auth {
+        tags[loan] = Lot(usr, wad);
+    }
 
-        uint wad = tag.price(loan);
+    function seize(uint loan) public {
+        (uint debt,,) = pile.loans(loan); // TODO: call debt registry or similar
+        require((threshold.get(loan) >= debt), "threshold-not-reached");
```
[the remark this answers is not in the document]
ilinzweilin: what about the comparison operator shouldn't it be the opposite way

manuelpolzhofer:
I think we can close this PR?
Changes are applied in @ilinzweilin  PR, correct?

ilinzweilin:
@manuelpolzhofer yes, my PR is based on this one. We can close it

PR #146, opened 2019-12-31
feat: trancheManager and Distributor

manuelpolzhofer:
- implemented trancheManager and Distributor with two tranches


src/core/tranche/manager.sol
```diff
@@ -0,0 +1,109 @@
...
+    function rely(address usr) public auth note { wards[usr] = 1; }
+    function deny(address usr) public auth note { wards[usr] = 0; }
+    modifier auth { require(wards[msg.sender] == 1); _; }
+
+    // --- Data ---
+    DistributorLike public distributor;
+    PileLike public pile;
+
+    // --- Tranches ---
+    OperatorLike public senior;
```
ilinzweilin: can the junior & senior vars not just be of type address

src/core/tranche/distributor/distributor.sol
```diff
@@ -0,0 +1,114 @@
...
+        if (requestCurrency > 0) {
+            requestCurrency = borrow(manager.senior(), requestCurrency);
+            return;
+        }
+        revert("request amount too high");
+    }
+
+    function borrow(address tranche, uint requestCurrency) internal returns(uint left) {
+        OperatorLike tranche = OperatorLike(tranche);
+        uint maxTranche = tranche.balance();
```
ilinzweilin: maybe better maxTake

src/core/tranche/distributor/distributor.sol
```diff
@@ -0,0 +1,114 @@
...
+
+    // -- Borrow Tranches ---
+    function borrowTranches(uint requestCurrency) internal  {
+        requestCurrency = borrow(manager.junior(), requestCurrency);
+
+        if (requestCurrency > 0) {
+            requestCurrency = borrow(manager.senior(), requestCurrency);
+            return;
+        }
+        revert("request amount too high");
```
ilinzweilin: From my understanding if requestCurrency is 0 then the borrow request was entirely covered by the junior tranche, so the transaction should not revert.

And also what if the borrow request can not be entirely covered by both tranches, should the transaction then revert? If yes we need to add an additional check after borrowing from senior

src/core/tranche/distributor/distributor.sol
ilinzweilin: what about computing maxTake based on the equityRatio?

src/core/tranche/distributor/distributor.sol
```diff
@@ -0,0 +1,114 @@
...
+        if (availableCurrency > 0) {
+            // junior gets the rest
+            OperatorLike(manager.junior()).repay(manager.pile(), availableCurrency);
+        }
+    }
+
+    function repay(address tranche, uint availableCurrency) internal returns(uint left) {
+        OperatorLike tranche = OperatorLike(tranche);
+        uint give = tranche.debt();
+        if (availableCurrency < tranche.debt()) {
```
ilinzweilin: < give
calling the var debt is maybe more descriptive.

src/core/tranche/distributor/trad_distributor.sol
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    function balance() public {
+        if(manager.poolClosing() == true) {
+            uint give = currency.balanceOf(manager.pile());
+            repayTranches(give);
+            return;
+        }
+
+        uint take = OperatorLike(manager.senior()).balance() + OperatorLike(manager.junior()).balance();
```
ilinzweilin: shouldn't we also consider the equityRation here

ilinzweilin:
Loooks good overall. Left some comments

src/core/tranche/manager.sol
manuelpolzhofer: yes, we could.  we might rename it to `seniorOperator` to highlight it is the operator contract address.

src/core/tranche/distributor/distributor.sol
manuelpolzhofer: Let's discuss it with the others. I will add the revert if.

src/core/tranche/distributor/trad_distributor.sol
manuelpolzhofer: It matters for the other distributor but it would be handled by the adapter not take to much.

src/core/tranche/distributor/distributor.sol
manuelpolzhofer: I think it would be better in the adapter to take care of the maxTake.

From the distributor perspective, it is okay to take the currency if it is already in the reserve

src/core/tranche/manager.sol
```diff
@@ -0,0 +1,109 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+pragma solidity >=0.4.24;
+
+import "ds-note/note.sol";
+import "../lightswitch.sol";
```
charleenfei: i think we don't need this import anymore :)

src/core/tranche/manager.sol
```diff
@@ -0,0 +1,109 @@
...
+
+    // --- Data ---
+    DistributorLike public distributor;
+    PileLike public pile;
+
+    // --- Tranches ---
+    OperatorLike public senior;
+    OperatorLike public junior;
+
+    // dominated in RAD
```
charleenfei: denominated

src/core/tranche/manager.sol
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    function file(bytes32 what, uint ratio) public auth {
+        if (what == "juniorRatio") { juniorRatio = ratio; }
+        else revert();
+    }
+
+    // --- Calls ---
+    function setTranche(bytes32 tranche, address operator_) public auth {
+        if (tranche == "junior") { junior = OperatorLike(operator_); }
```
charleenfei: maybe we could put this into the `depend` method?

src/core/tranche/distributor/distributor.sol
charleenfei: I thought we talked about using `borrow and repay` bc `give and take` depends on the perspective of which side the distributor is engaging with

src/core/tranche/distributor/distributor.sol
```diff
@@ -0,0 +1,117 @@
...
+    // -- Borrow Tranches ---
+    function borrowTranches(uint requestCurrency) internal  {
+        requestCurrency = borrow(manager.junior(), requestCurrency);
+
+        if (requestCurrency > 0) {
+            requestCurrency = borrow(manager.senior(), requestCurrency);
+            return;
+        }
+
+        if (requestCurrency > 0) {
```
charleenfei: ? also, this if condition will never be fulfilled the way these statements are ordered

src/core/tranche/manager.sol
```diff
@@ -0,0 +1,109 @@
...
+    function deny(address usr) public auth note { wards[usr] = 0; }
+    modifier auth { require(wards[msg.sender] == 1); _; }
+
+    // --- Data ---
+    DistributorLike public distributor;
+    PileLike public pile;
+
+    // --- Tranches ---
+    OperatorLike public senior;
+    OperatorLike public junior;
```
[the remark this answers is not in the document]
charleenfei: i think we should use junior, have confirmed with Jason this is the more specific term

src/core/tranche/distributor/distributor.sol
```diff
@@ -0,0 +1,117 @@
...
+            revert("requested currency amount too high");
+        }
+    }
+
+    function borrow(address tranche, uint requestCurrency) internal returns(uint left) {
+        OperatorLike tranche = OperatorLike(tranche);
+        uint maxBorrow = tranche.balance();
+        uint borrowAmount = maxBorrow;
```
charleenfei: i think you could simplify this and remove maxBorrow, just use `borrowAmount = tranche.balance()`

src/core/tranche/distributor/distributor.sol
manuelpolzhofer: changed

src/core/tranche/manager.sol
manuelpolzhofer: removed

PR #147, opened 2019-12-31
refactor quant and operator

ilinzweilin:
- Removed iSupply dependency from the quant.
- Introduced base operator contract and extended senior operator that uses the quant to keep track of the tranche debt. 
- Fixed proxy tests: loan ids now start at 1 instead of 0 (required for the multiple loans per NFT fix)


src/core/tranche/seniorOperator.sol
```diff
@@ -0,0 +1,43 @@
...
+
+contract QuantLike {
+   function updateDebt(int) public;
+}
+
+// SeniorOperator
+// Interface to the senior tranche. Uses the quant to keep track of the current debt towards the tranche. 
+contract SeniorOperator is Operator {
+
+    QuantLike public quant;
```
lucasvo: Couldn't we inline this contract here?

src/core/tranche/operator.sol
```diff
@@ -26,18 +26,13 @@ contract ReserveLike{
...
-
 contract SlicerLike {
    function getSlice(uint) public returns(uint);
    function getPayout(uint) public returns(uint);
 }
 
 // Operator
-// Manages the reserve. Triggers supplyRate & borrowRate calculations.
+// Interface of a tranche. Coordinates investments and borrows to/from the tranche.
 contract Operator is DSNote {
```
lucasvo: If we don't use the note modifier anywhere, we shouldn't extend from DSNote

ilinzweilin: isn't it used with all the public functions

src/core/tranche/seniorOperator.sol
```diff
@@ -15,23 +15,11 @@
...
-    // --- Auth ---
-    mapping (address => uint) public wards;
-    function rely(address usr) public auth note { wards[usr] = 1; }
-    function deny(address usr) public auth note { wards[usr] = 0; }
-    modifier auth { require(wards[msg.sender] == 1); _; }
-
-    // --- Data ---
-    ReserveLike public reserve;
+// SeniorOperator
+// Interface to the senior tranche. Uses the quant to keep track of the current debt towards the tranche.
```
charleenfei: I think we need to edit this comment, it still refers to the quant

src/core/tranche/seniorOperator.sol
```diff
@@ -41,44 +29,27 @@ contract Quant is DSNote {
...
-        }
     }
 
-    function file(bytes32 what, bool data) public note auth {
-         if (what ==  "fixedsupplyrate") {
-             supplyRateFixed = data;
-        }
+    function repay(address usr, uint currencyAmount) public note auth {
+        super.repay(usr, currencyAmount);
+        updateDebt(int(currencyAmount) * -1);
```
charleenfei: shouldn't this be `int(currencyAmount * -1)`?

src/core/tranche/test/operator.t.sol
```diff
@@ -65,33 +66,24 @@ contract OperatorTest is DSTest {
...
         operator.borrow(address(this), currencyAmount);
         
         assertEq(reserve.callsBorrow(), 1);
         assertEq(reserve.usr(), address(this));
         assertEq(reserve.currencyAmount(), currencyAmount);
-        assertEq(quant.callsUpdateDebt(), 1);
-        assertEq(quant.loanAmount(), int(currencyAmount));
-        assertEq(quant.callsUpdateBorrowRate(), 1);
+        assertEq(operator.debt(), currencyAmount);
+        emit log_named_uint("moin", operator.debt());
```
charleenfei: moin!

src/core/tranche/seniorOperator.sol
charleenfei: also maybe we should updateDebt before repay (reentrancy)

src/core/tranche/seniorOperator.sol
```diff
@@ -41,44 +29,27 @@ contract Quant is DSNote {
...
-                drip();
-            }
-            uint balance = reserve.balance();
-            uint supplyRate_ = sub(supplyRate, ONE);
-            uint ratio = rdiv(add(balance, debt), debt); 
-            borrowRate.speed = add(rmul(ratio, supplyRate_), ONE);
-        }
+    function borrow(address usr, uint currencyAmount) public note auth {
+        super.borrow(usr, currencyAmount);
+        updateDebt(int(currencyAmount));
```
charleenfei: see re-entrancy possibility above

src/core/tranche/seniorOperator.sol
ilinzweilin: I think if I cast the currencyAmount to int then the operation is already performed in the int domain. But in general we could split it in increase and decrease debt to avoid type casts....

PR #156, opened 2020-01-09
feat: add operator contract (old name bouncer)

manuelpolzhofer:
- I had a discussion with Lucas
- we decided the bouncer contract should fetch the token price from the assessor
- it would allow the admin to give customized prices if required
- we need a different auth pattern because otherwise, users could add  other users with `rely`

- changed name of bouncer to operator


src/core/lender/tranche/operator.sol
```diff
@@ -0,0 +1,66 @@
...
+        assessor = AssessorLike(assessor_);
+    }
+
+    function depend(bytes32 what, address addr) public auth {
+        if (what == "tranche") { tranche = TrancheLike(addr); }
+        else if (what == "assessor") { assessor = AssessorLike(addr); }
+        else revert();
+    }
+
+    function supply(uint currencyAmount) public auth_external {
```
ilinzweilin: aren't we supposed to check here, if supply is active

src/core/lender/tranche/operator.sol
```diff
@@ -0,0 +1,66 @@
...
+        tranche.supply(msg.sender, currencyAmount, rdiv(currencyAmount, assessor.calcTokenPrice()));
+
+    }
+
+    function redeem(uint tokenAmount) public auth_external {
```
ilinzweilin: same here with redeeem

manuelpolzhofer:
I will merge the PR and address the flags in a seperated PR

PR #157, opened 2020-01-09
chore: assessor calcTokenPrice and tests

manuelpolzhofer:
- added tests for assessor
- handle edge cases for calcTokenPrice
- simpler tranche mock


src/core/lender/test/mock/tranche.sol
```diff
@@ -12,43 +11,24 @@
...
-    uint public rate;
-    uint public reserve;
-    int public loanAmount;
-
+    mapping (bytes32 => uint) calls;
+    mapping (bytes32 => uint) returnValues;
+    function setReturn(bytes32 name, uint returnValue) public {
+        returnValues[name] = returnValue;
+    }
+    function funcBody(bytes32 name) internal returns (uint) {
```
charleenfei: 👍 this is cool, we should introduce this pattern for all of our mocks

src/core/lender/assessor.sol
```diff
@@ -58,19 +61,25 @@ contract Assessor is DSNote,DSMath {
         else revert();
     }
 
+    function calcAssetValue(address tranche) public returns(uint) {
+        uint trancheReserve = TrancheLike(tranche).balance();
+        uint poolValue = pile.debt();
+        if (tranche == junior) {
+            return calcJuniorAssetValue(poolValue, trancheReserve, seniorDebt());
+        }
+        return calcSeniorAssetValue(poolValue, trancheReserve, SeniorTrancheLike(tranche).debt(), juniorReserve());
```
charleenfei: you could just used seniorDebt() instead of SeniorTrancheLike(tranche).debt()

src/core/lender/test/mock/tranche.sol
ilinzweilin: sounds good. How about calling the function simply -> call()

src/core/lender/assessor.sol
```diff
@@ -58,19 +61,25 @@ contract Assessor is DSNote,DSMath {
...
-        } else {
-            assetValue =  calcSeniorAssetValue(poolValue, trancheReserve, SeniorTrancheLike(tranche_).debt(), juniorReserve());
+        uint tokenSupply = TrancheLike(msg.sender).tokenSupply();
+        uint assetValue = calcAssetValue(msg.sender);
+        if (tokenSupply == 0) {
+            return ONE;
         }
-
-        return rdiv(assetValue, TrancheLike(tranche_).tokenSupply());
+        if (assetValue == 0) {
```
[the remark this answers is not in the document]
manuelpolzhofer: if `tokenSupply == 0` it is okay for `assetValue == 0`  because it is the inital setup

src/core/lender/test/assessor.t.sol
```diff
@@ -1,102 +1,126 @@
...
+        tokenSupply = 2.7182818284590452 ether;
+        debt = 3.14159265359 ether;
+        senior.setReturn("debt", debt);
+        pile.setDebtReturn(debt);
+        assetValue = assessor.calcAssetValue(address(senior));
+        // sanity check
+        assertEq(assetValue, debt);
+        senior.setReturn("tokenSupply",tokenSupply);
+        tokenPrice = senior.doCalcTokenPrice(assessor_);
+        assertEq(tokenPrice, rdiv(assetValue, tokenSupply));
```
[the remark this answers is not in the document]
manuelpolzhofer: yes. also for token price with junior tranche

src/core/lender/assessor.sol
manuelpolzhofer: we could but I think it is a bit cheaper not doing it

src/core/lender/assessor.sol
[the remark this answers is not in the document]
manuelpolzhofer: if `tokenSupply == 0 ` we can always return `ONE`

PR #167, opened 2020-01-10
refactor: borrower contracts

ilinzweilin:
This PR is based on @lucasvo Collector Refactor branch.

**Changes:** 

- Collector logic: remove Spotter & Tag
- Shelf: Only one loan can be whitelisted per NFT, remove principal management
- Merge Pile & Shelf
- DebtRegister: add rate management, store debt values in pie format, use Maker naming style, rename to Pile 
- Ceiling (Manages the max amount that can be borrowed): added two contracts Principal & CreditLine

**Consequences:**

Admin: 
- remove dependency for shelf (prev pile)
- use new pile interface (prev debtRegister) 

Admit:
- use directly a contract implementing the ceiling interface to set & update loan ceiling, like principal
- use new shelf interface

Deployer:
- added PrincipalFab 

Mocks created/adjusted for:
- Pile 
- Shelf
- Ceiling = Principal/CreditLine

Tests added/adjusted for:
- principal
- creditline
- system
- admin
- admit
- collector
- pile
- shelf

Open questions/Todos:
- decrease balance when loan is recovered
- introducing a check on the contract level if rate group has already a speed set before calling file in pile. The method always resets rho & chi
- in collector: current logic allows a loan to be collected by anyone when no keeper is set for the loan --> address(0). 
  Maybe it would make sence to also check if the price is above 0. 
The PR is already really 
- Check the criteria that checks if a loan can be collected.
- There are still some Todos regarding reentrancy & deployer logic (marked in the contracts). Since the PR is already really huge, would be good to create tickets and address in following prs.


src/core/admin.sol
```diff
@@ -72,22 +67,22 @@ contract Admin {
         uint loan = admit.admit(registry, nft, principal, usr);
         appraiser.file(loan, appraisal);
 
-        (,, uint speed,) = debtRegister.rates(rate);
+        (,, uint speed,) = pile.rates(rate);
```
lucasvo: the whitelist method should be removed

src/core/ceiling/creditline.sol
```diff
@@ -0,0 +1,65 @@
+// store.sol
```
lucasvo:
```suggestion
// cerditline.sol
```

src/core/ceiling/creditline.sol
```diff
@@ -0,0 +1,65 @@
...
+    function depend(bytes32 what, address addr) public auth {
+        if (what == "pile") { pile = PileLike(addr); }
+        else revert();
+    }
+
+    function file(uint loan, uint creditLine) public note auth {
+        values[loan] = creditLine;
+    }
+
+    function borrow(uint loan, uint currencyAmount) public note auth {
```
lucasvo:
```suggestion
    function borrow(uint loan, uint currencyAmount) public auth {
```

src/core/ceiling/creditline.sol
```diff
@@ -0,0 +1,65 @@
...
+        require(values[loan] >= add(pile.debt(loan), currencyAmount));
+    }
+
+    function add(uint x, uint y) internal pure returns (uint z) {
```
lucasvo: use tinlake-math?

src/core/ceiling/principal.sol
```diff
@@ -0,0 +1,54 @@
+// store.sol
```
lucasvo:
```suggestion
// principal.sol
```

src/core/ceiling/principal.sol
```diff
@@ -0,0 +1,54 @@
...
+    function file(uint loan, uint principal) public note auth {
+        values[loan] = principal;
+    }
+
+    function borrow(uint loan, uint amount) public note auth {
+        require(values[loan] >= amount);
+        values[loan] = sub(values[loan], amount);
+    }
+
+    function sub(uint x, uint y) internal pure returns (uint z) {
```
lucasvo: use tinlake-math

src/core/ceiling/principal.sol
```diff
@@ -0,0 +1,54 @@
...
+
+    function ceiling(uint loan) public returns(uint) {
+        return values[loan];
+    }
+
+    function file(uint loan, uint principal) public note auth {
+        values[loan] = principal;
+    }
+
+    function borrow(uint loan, uint amount) public note auth {
```
lucasvo: presumably there's already a note on shelf.borrow() so this should not be needed
```suggestion
    function borrow(uint loan, uint amount) public auth {
```

src/core/collect/collector.sol
```diff
@@ -16,62 +16,83 @@
...
+        uint debt = pile.debt(loan);
+        require((threshold.get(loan) <= debt), "threshold-not-reached");
+        shelf.claim(loan, address(this));
+    }
 
-        pile.recovery(loan, msg.sender, wad);
-        spotter.free(loan, usr);
+    function collect(uint loan) public auth {
+        require(msg.sender == tags[loan].usr || tags[loan].usr == address(0));
+        // TODO: reentrancy?
```
lucasvo:
```suggestion
```

src/core/collect/test/collector.t.sol
```diff
@@ -0,0 +1,122 @@
...
+        uint debt = 100;
+        uint price = debt-1;
+        setUpLoan(loan, tokenId, debt);
+
+        threshold.set(loan, debt+1);
+        collector.file(loan, address(1), price);
+        seize(loan);
+        collect(loan, tokenId, price);
+    }
+    // TODO: Missing test cases
```
lucasvo: can be removed

src/core/deployer.sol
```diff
@@ -68,93 +62,101 @@ contract LightSwitchFab {
...
-        shelf = new Shelf(pile, appraiser, title);
+   function newShelf(address tkn_, address title_, address debt_, address Principal_) public returns (Shelf shelf) {
+        shelf = new Shelf(tkn_, title_, debt_, Principal_);
         shelf.rely(msg.sender);
         shelf.deny(address(this));
     }
 }
 
 contract DeskFab {
     // note: this is the mock Desk, which will interface with the lender/tranche side of Tinlake, and does not require auth for now.
```
lucasvo: this note is weird, is it still relevant?

src/core/deployer.sol
```diff
@@ -68,93 +62,101 @@ contract LightSwitchFab {
...
         adminfab = adminfab_;
-        debtRegisterfab = debtRegisterfab_;
+        pilefab = pilefab_;
+        principalFab = principalFab_;
     }
 
-    function deployCollect(address collectDeployer_ ,uint threshold_) public {
-        collectDeployer = CollectDeployerLike(collectDeployer_);
-        collectDeployer.deploy(address(pile), address(shelf), address(desk), threshold_);
+    function deployCollect(address collectDeployer_ , uint threshold_) public {
```
lucasvo: looks like it can be removed?

src/core/pile.sol
```diff
@@ -17,150 +17,213 @@ pragma solidity >=0.4.24;
...
-    function decLoanDebt(uint, uint, uint) public;
-    function drip(uint) public;
-}
-
-// Pile
-// Manages the balance for the currency ERC20 in which borrowers want to borrow.
-contract Pile is DSNote, TitleOwned {
+// ## Interst Group based Pile
+// The following is one implementation of a debt module. It keeps track of different buckets of interest rates and is optimized for many loans per interest bucket. It keeps track of interest
+// rate accumulators (chi values) for all interest rate categories. It Calculates debt each
```
lucasvo:
```suggestion
// rate accumulators (chi values) for all interest rate categories. It calculates debt each
```

src/core/shelf.sol
```diff
@@ -17,61 +17,61 @@
...
     // --- Auth ---
     mapping (address => uint) public wards;
     function rely(address usr) public auth { wards[usr] = 1; }
     function deny(address usr) public auth { wards[usr] = 0; }
     modifier auth { require(wards[msg.sender] == 1); _; }
 
     // --- Data ---
-    PileLike                  public pile;
-    AppraiserLike             public appraiser;
+    Title public title;
```
lucasvo: TitleLike I think is enough, no?

src/core/shelf.sol
```diff
@@ -17,61 +17,61 @@
 pragma solidity >=0.4.24;
 pragma experimental ABIEncoderV2;
 
-import { TitleOwned } from "tinlake-title/title.sol";
-
-contract AppraiserLike {
-    function appraise(uint, address, uint) public returns (uint);
-}
+import "ds-note/note.sol";
+import { Title, TitleOwned } from "tinlake-title/title.sol";
```
lucasvo:
```suggestion
import { TitleLike, TitleOwned } from "tinlake-title/title.sol";
```

src/core/pile.sol
```diff
@@ -17,150 +17,213 @@ pragma solidity >=0.4.24;
...
+        drip(rate);
+    }
+
+    // accrue neeeds to be called before any debt amounts are modified by an external component
+    function accrue(uint loan) public {
+        drip(group[loan]);
+    }
+    
+    // compound calculates the new chi, the delta between old and new chi and the delta of debt
+    function compound(uint rate) internal view returns (uint chi, uint delta) {
```
lucasvo: I did not review this because it will be moved to tinlake-math

src/core/shelf.sol
```diff
@@ -17,61 +17,61 @@
...
-        appraiser = AppraiserLike(appraiser_);
+        ceiling = CeilingLike(ceiling_);
     }
 
-    // --- Shelf ---
-
     function depend (bytes32 what, address addr) public auth {
-        if (what == "pile") { pile = PileLike(addr); }
-        else if (what == "appraiser") { appraiser = AppraiserLike(addr); }
+        if (what == "lender") { lender = addr; }
```
lucasvo: We should add all the other components in here as well (title, pile ceiling)

src/core/shelf.sol
```diff
@@ -80,54 +80,104 @@ contract Shelf is TitleOwned {
...
+        shelf[loan].tokenId = token;
+
+        return loan;
     }
 
-    // Move the NFT out of the shelf
-    function move(uint loan, address registry_, uint nft_, address to) public owner(loan) {
-        NFTLike(registry_).transferFrom(address(this), to, nft_);
+    function close(uint loan) public owner(loan) {
+        require(pile.debt(loan) == 0, "outstanding-debt"); // TODO: only allow closing of a loan that isn't active anymore. maybe there is a better criteria
```
lucasvo: @ilinzweilin did you discuss with the team if this criteria is safe? It does lead to an edge case where if a loan is collected, the NFT can never be used again. Perhaps we need to allow the collector to close a loan without that check?

src/core/shelf.sol
```diff
@@ -80,54 +80,104 @@ contract Shelf is TitleOwned {
...
     }
 
-    function release (uint loan, address usr) public owner(loan) {
-        uint debt = pile.debtOf(loan);
-        require(debt == 0, "debt");
-        move(loan, shelf[loan].registry, shelf[loan].tokenId, usr);
-        adjust(loan);
+    // --- Currency actions ---
+    function want() public view returns (int) {
+        return int(balance) - int(tkn.balanceOf(address(this))); // safemath
```
lucasvo: This should not return an int but a boolean flag and a value.

src/core/shelf.sol
```diff
@@ -80,54 +80,104 @@ contract Shelf is TitleOwned {
...
+
+    function repay(uint loan, uint wad) public owner(loan) note {
+        require(nftLocked(loan), "nft-not-locked");
+        require(balances[loan] == 0,"before repay loan needs to be withdrawn");
+        doRepay(loan, msg.sender, wad);
+    }
+
+    function recover(uint loan, address usr, uint wad) public auth {
+        doRepay(loan, usr, wad);
+        // TODO: do we need to decrease the loan balance here? There might be some
```
lucasvo: @ilinzweilin what about this one?

src/core/shelf.sol
```diff
@@ -80,54 +80,104 @@ contract Shelf is TitleOwned {
...
-        NFTLike(shelf[loan].registry).transferFrom(address(this), usr, shelf[loan].tokenId);
-        adjust(loan);
+    function borrow(uint loan, uint wad) public owner(loan) {
+        require(nftLocked(loan), "nft-not-locked");
+        ceiling.borrow(loan, wad);
+        pile.accrue(loan);
+        pile.incDebt(loan, wad);
+
+        balances[loan] = add(balances[loan], wad);
+        balance += wad;
```
lucasvo:
```suggestion
        balance = add(balance, wad);
```

src/core/shelf.sol
```diff
@@ -80,54 +80,104 @@ contract Shelf is TitleOwned {
...
     }
 
-    function deposit (uint loan, address usr) public owner(loan) {
+    function withdraw(uint loan, uint wad, address usr) public owner(loan) note {
+        require(nftLocked(loan), "nft-not-locked");
+        require(wad <= balances[loan], "amount-too-high");
+        balances[loan] -= wad;
```
lucasvo: safe math pls

src/core/shelf.sol
```diff
@@ -80,54 +80,104 @@ contract Shelf is TitleOwned {
...
+    }
+
+    function recover(uint loan, address usr, uint wad) public auth {
+        doRepay(loan, usr, wad);
+        // TODO: do we need to decrease the loan balance here? There might be some
+        uint loss = pile.debt(loan);
+        pile.decDebt(loan, loss);
+    }
+
+    function doRepay(uint loan, address usr, uint wad) internal {
```
lucasvo: I would propose we adopt the openzeppelin convention here: prefixing `_` to internal methods:
```suggestion
    function _repay(uint loan, address usr, uint wad) internal {
```

src/core/shelf.sol
```diff
@@ -80,54 +80,104 @@ contract Shelf is TitleOwned {
...
+            wad = loanDebt;
+        } 
+
+        tkn.transferFrom(usr, address(this), wad);
+        pile.decDebt(loan, wad);
+        tkn.approve(lender, wad);
+    }
+
+    // --- NFT actions ---
+    // deposit = lock
```
lucasvo:
```suggestion
```

src/core/shelf.sol
```diff
@@ -80,54 +80,104 @@ contract Shelf is TitleOwned {
...
+        require(pile.debt(loan) == 0, "has-debt");
+        NFTLike(shelf[loan].registry).transferFrom(address(this), msg.sender, shelf[loan].tokenId);
+    }
+
+    function nftLocked(uint loan) internal returns (bool) {
+        return NFTLike(shelf[loan].registry).ownerOf(shelf[loan].tokenId) == address(this);
+    }
+
+    function claim(uint loan, address usr) public auth {
+        // TODO: need to update pile/shelf to let it know it's gone.
```
lucasvo: @ilinzweilin

src/core/admin.sol
ilinzweilin: 👍https://github.com/centrifuge/tinlake/issues/168

src/core/ceiling/creditline.sol
ilinzweilin: 👍 https://github.com/centrifuge/tinlake/issues/169

src/core/deployer.sol
ilinzweilin: we will remove this it anyway when integrating the new borrower & lender contracts

src/core/shelf.sol
ilinzweilin: TitleLike needs to be extended with issue & close function signatures first

src/core/shelf.sol
ilinzweilin: Created a research/discussion ticket and added some topics that came up. @lucasvo @charly @Manuel please add topics. We discuss on Monday https://github.com/centrifuge/tinlake/issues/170

src/core/shelf.sol
ilinzweilin: https://github.com/centrifuge/tinlake/issues/171

src/core/shelf.sol
ilinzweilin: Just created a refactoring/styling ticket to collect all the changes we want to implement for v2. https://github.com/centrifuge/tinlake/issues/172
@lucasvo @charleenfei @manuelpolzhofer let's add all the ideas/request there

src/core/pile.sol
```diff
@@ -17,150 +17,213 @@ pragma solidity >=0.4.24;
...
-        debtRegister = DebtRegisterLike(debtRegister_);
+        rates[0].chi      = ONE;
+        rates[0].speed    = ONE;
+    }
+
+    // --- Public Debt Methods  ---
+    // increase the loan's debt by wad
+    function incDebt(uint loan, uint wad) public auth note {
+        uint rate = group[loan];
+        require(now <= rates[rate].rho);
```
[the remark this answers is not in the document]
ilinzweilin: I think @lucasvo introduced it in consideration of potential EVM block timestamp behavior changes

src/core/pile.sol
```diff
@@ -17,150 +17,213 @@ pragma solidity >=0.4.24;
...
 
-import { TitleOwned } from "tinlake-title/title.sol";
 
-contract TokenLike{
-    uint public totalSupply;
-    function balanceOf(address) public view returns (uint);
-    function transferFrom(address,address,uint) public;
-    function approve(address, uint) public;
+// ## PileLike the interface for any debt module.
+contract PileLike {
```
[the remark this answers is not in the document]
ilinzweilin: Moved it out of pile for consistency. But it would make sense to move the interfaces into sep files. They should be independent of the contracts implementing, but always adding them to contracts using them generates a lot of duplicated code 🙈

src/core/shelf.sol
lucasvo: It's much harder to find these kind of things later on. I really would recommend to just fix these things right away.

src/core/pile.sol
```diff
@@ -17,150 +17,203 @@ pragma solidity >=0.4.24;
...
-    function initRate(uint, uint) public;
-    function incLoanDebt(uint, uint, uint) public;
-    function decLoanDebt(uint, uint, uint) public;
-    function drip(uint) public;
-}
-
-// Pile
-// Manages the balance for the currency ERC20 in which borrowers want to borrow.
-contract Pile is DSNote, TitleOwned {
+// ## Interst Group based Pile
```
charleenfei: Interest

src/core/test/mock/ceiling.sol
```diff
@@ -0,0 +1,31 @@
+pragma solidity >=0.4.24;
+
+contract CeilingMock {
+    uint public callsBorrow;
+    uint public callsFile;
+    bool public celeingReached; function setCeilingReached(bool celeingReached_) public {celeingReached=celeingReached_;}
```
charleenfei: ceiling 🙈

src/core/ceiling/principal.sol
```diff
@@ -0,0 +1,57 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+pragma solidity >=0.4.24;
+pragma experimental ABIEncoderV2;
+
+import "ds-note/note.sol";
+
+// Principal is an implementation of the Ceiling module that defines the max amoutn a user can borrow.
```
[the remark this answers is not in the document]
ilinzweilin: they implement different logic

src/core/test/mock/ceiling.sol
ilinzweilin: 🦅 👀

src/core/ceiling/principal.sol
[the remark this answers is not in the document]
charleenfei: got it 👍

src/core/ceiling/creditline.sol
```diff
@@ -0,0 +1,75 @@
...
+
+    function file(uint loan, uint creditLine) public auth {
+        values[loan] = creditLine;
+    }
+
+    function borrow(uint loan, uint amount) public auth {
+        require(values[loan] >= add(pile.debt(loan), amount));
+    }
+
+    function repay(uint loan, uint amount) public note auth {
```
manuelpolzhofer: lets add a todo here

src/core/collect/collector.sol
```diff
@@ -16,62 +16,83 @@
...
 
-        pile.recovery(loan, msg.sender, wad);
-        spotter.free(loan, usr);
+    function collect(uint loan, address usr) public auth {
+        require(usr == tags[loan].usr || tags[loan].usr == address(0));
+        // TODO: reentrancy?
+        (address registry, uint nft) = shelf.token(loan);
+        NFTLike(registry).transferFrom(address(this), usr, nft);
+        shelf.recover(loan, usr, tags[loan].wad);
         desk.balance();
```
manuelpolzhofer: lets not call it desk here

src/core/pile.sol
```diff
@@ -17,150 +17,203 @@ pragma solidity >=0.4.24;
...
+        uint debt = fromPie(rates[currentRate].chi, pie_);
+        rates[currentRate].pie = sub(rates[currentRate].pie, pie_);
+        pie[loan] = toPie(rates[newRate].chi, debt);
+        rates[newRate].pie = add(rates[newRate].pie, pie[loan]);
+        group[loan] = newRate;
+    }
+
+    // set/change the interest rate of a rate category
+    function file(uint rate, uint speed_) public auth {
+        require(speed_ != 0);
```
manuelpolzhofer: we need `rates[rate].chi == 0` here because changing the speed already in use rate group would work differently.

It would just update the `speed` and not re-init the `chi`

src/core/pile.sol
```diff
@@ -17,150 +17,203 @@ pragma solidity >=0.4.24;
...
+        rates[rate].speed = speed_;
+        rates[rate].chi = ONE;
+        rates[rate].rho = uint48(now);
+        drip(rate);
```
manuelpolzhofer: I think the `drip` call is not needed

src/core/pile.sol
```diff
@@ -17,150 +17,203 @@ pragma solidity >=0.4.24;
...
+            rates[rate].rho = uint48(now);
+            total = add(total, delta);
+        }
     }
 
-    function depend(bytes32 what, address data) public auth {
-        if (what == "lender") { lender = data; }
+    // convert debt amount to pie
+    function toPie(uint chi, uint pie) internal view returns (uint) {
+        return rdiv(pie, chi);
```
manuelpolzhofer: We need to use the rounding div method here

src/core/shelf.sol
```diff
@@ -80,54 +91,105 @@ contract Shelf is TitleOwned {
...
-        shelf[loan].initial = principal_;
+    function token(uint loan) public view returns (address registry, uint nft) {
+        return (shelf[loan].registry, shelf[loan].tokenId);
     }
 
-    function file(uint loan, uint principal_) public auth {
-        shelf[loan].principal = principal_;
-        shelf[loan].initial = principal_;
+    // --- Shelf: Loan actions ---
+    function issue(address registry, uint token) public returns (uint) {
```
manuelpolzhofer: Currently, the method is public. I don't know if we should have an auth here with a wrapper contract which allows every address to pass but we have the option to block it.

Everybody could just spam our loan issue method.  (Which might be a problem in the UI)

src/core/shelf.sol
```diff
@@ -80,54 +91,105 @@ contract Shelf is TitleOwned {
...
+
+        return loan;
     }
 
-    // Move the NFT out of the shelf
-    function move(uint loan, address registry_, uint nft_, address to) public owner(loan) {
-        NFTLike(registry_).transferFrom(address(this), to, nft_);
+    function close(uint loan) public owner(loan) {
+        require(pile.debt(loan) == 0, "outstanding-debt");
+        title.close(loan);
```
manuelpolzhofer: Why are we doing this? It requires an additional `approve` from the loan owner

src/core/shelf.sol
```diff
@@ -80,54 +91,105 @@ contract Shelf is TitleOwned {
...
+        nftlookup[nft] = 0;
     }
 
-    function release (uint loan, address usr) public owner(loan) {
-        uint debt = pile.debtOf(loan);
-        require(debt == 0, "debt");
-        move(loan, shelf[loan].registry, shelf[loan].tokenId, usr);
-        adjust(loan);
+    // --- Currency actions ---
+    function want() public view returns (int) {
```
manuelpolzhofer: I guess this will be refactored to the new interface :)

src/core/shelf.sol
```diff
@@ -80,54 +91,105 @@ contract Shelf is TitleOwned {
...
+        return int(balance) - int(tkn.balanceOf(address(this)));
     }
 
-    function free(uint loan, address usr) public auth  {
-        NFTLike(shelf[loan].registry).transferFrom(address(this), usr, shelf[loan].tokenId);
-        adjust(loan);
+    function borrow(uint loan, uint wad) public owner(loan) {
+        require(nftLocked(loan), "nft-not-locked");
+        ceiling.borrow(loan, wad);
+        pile.accrue(loan);
```
manuelpolzhofer: I think it is not required to call `accure` here. It might be nice to have the debt of the rate up to date but not required

src/core/shelf.sol
```diff
@@ -80,54 +91,105 @@ contract Shelf is TitleOwned {
...
+        uint loanDebt = pile.debt(loan);
+        // only repay max loan debt
+        if (wad > loanDebt) {
+            wad = loanDebt;
+        } 
+
+        tkn.transferFrom(usr, address(this), wad);
+        pile.decDebt(loan, wad);
+        tkn.approve(lender, wad);
```
manuelpolzhofer: I guess we can remove it. The distributor will have full access to currency in pile

manuelpolzhofer: but lets discuss

src/core/shelf.sol
```diff
@@ -80,54 +91,105 @@ contract Shelf is TitleOwned {
...
+    }
+
+    // --- NFT actions ---
+    function lock(uint loan, address usr) public owner(loan) {
         NFTLike(shelf[loan].registry).transferFrom(usr, address(this), shelf[loan].tokenId);
```
manuelpolzhofer: I think this could be a security issue.

Lets say A and B both are using Tinlake and both approved their NFTs to be taken by the shelf.

A could transfer the NFT to B.
B approves shelf to take the NFT.
**A could afterwards still trigger the lock of the NFT**

what happen if we require loan owner and nft owner must always be the same by using `msg.sender` instead of `usr`

src/core/shelf.sol
ilinzweilin: k, integrated into this pr

src/core/collect/collector.sol
```diff
@@ -16,62 +16,83 @@
...
-        uint wad = tag.price(loan);
+    function seize(uint loan) public {
+        uint debt = pile.debt(loan);
+        require((threshold.get(loan) <= debt), "threshold-not-reached");
+        shelf.claim(loan, address(this));
+    }
 
-        pile.recovery(loan, msg.sender, wad);
-        spotter.free(loan, usr);
+    function collect(uint loan) public auth {
```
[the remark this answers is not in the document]
ilinzweilin: integrated: `
    function collect(uint loan, address usr) public auth {
        require(usr == tags[loan].usr || tags[loan].usr == address(0));
}`

src/core/pile.sol
```diff
@@ -17,150 +17,203 @@ pragma solidity >=0.4.24;
...
+        if (now >= rates[rate].rho) {
+            (chi,) = compound(rate);
+        } 
+        return fromPie(chi, pie);
+    } 
+
+    // --- Interest Rate Group Implementation ---
+
+    // set rate group for a loan
+    function setRate(uint loan, uint rate) public auth {
```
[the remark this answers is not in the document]
ilinzweilin: That is exactly one of the questions I added to the discussion ticket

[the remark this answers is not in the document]
ilinzweilin: added the check, as I also think it makes it safer to to mess up the rate group

src/core/admit.sol
```diff
@@ -37,43 +42,46 @@ contract Admit {
...
 
     function update(uint loan, address registry_, uint nft_, uint principal_) public auth {
-        (, , ,uint principal) = shelf.shelf(loan);
+        uint principal = ceiling.values(loan);
 
         // loan status should be whitelisted
-        require(principal != 0);
-        shelf.file(loan, registry_, nft_, principal_);
-
+        require(principal != 0, "loan not whitelisted");
```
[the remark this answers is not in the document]
lucasvo: I think we need to have the admit contract anymore in the future, we can get rid of it.

src/core/shelf.sol
lucasvo: No, title.close is an auth action and doesn't need an `approve`

src/core/shelf.sol
lucasvo: The order is incorrect. It should be:
```
        pile.accrue(loan);
        ceiling.borrow(loan, wad);
```

It should be in there like this so that the ceiling can be made dependent on the current debt amount. We should call pile.accrue at the beginning of borrows & repay.

src/core/shelf.sol
lucasvo: Yes, it might need to be `msg.sender`

src/core/shelf.sol
manuelpolzhofer: You are calling the `_burn` method of the ERC721 which requires ownership or approve. 
Ownership is still the borrower and not the shelf (because it is the title)

or I am getting something wrong?

src/core/shelf.sol
manuelpolzhofer: Good point

src/core/ceiling/principal.sol
```diff
@@ -0,0 +1,57 @@
...
+    constructor() public {
+        wards[msg.sender] = 1;
+    }
+
+    function ceiling(uint loan) public returns(uint) {
+        return values[loan];
+    }
+
+    function file(uint loan, uint principal) public note auth {
+        values[loan] = principal;
```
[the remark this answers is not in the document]
lucasvo: Yes it's fine. It's only used in borrow()

src/core/collect/collector.sol
lucasvo: @ilinzweilin pls fix

src/core/pile.sol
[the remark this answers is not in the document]
lucasvo: https://github.com/centrifuge/tinlake/pull/167/files/d8ccb527ceade2494e49e7b457f1d3cce130eca7#diff-e46d09b7a05c83ae1a2a8d09b1b9a8b6R52 

In this line we set the 0 rate group to just mean no interest. But with this change we imply that the rategroup is not set at all if it's 0. I think that will lead to issues. Do we want to forbid setting it to 0? Then we should fix the above code (either by deleting it because it's not needed or by moving it to rate group 1?)

src/core/pile.sol
lucasvo: Proposal for a safer implementation that calls drip before setting the rate and sets chi to 1 if chi is 0 (i.e. the group is created)

```
    function file(uint rate, uint speed_) public auth {
        require(speed_ != 0);

        if (rates[chi] == 0) { 
            rates[rate].chi = ONE;
            rates[rate].rho = unit48(now);
        } else { 
            drip(rate);
        }

        rates[rate].speed = speed_; 
}

File: src/core/pile.sol (hunk 34 above)
Comment (lucasvo) [reply src/core/pile.sol]: @manuelpolzhofer is that correct in tinlake-math? I'd strongly vote for refactoring this right now instead of reviewing this code here only to replace it a day or two later.

File: src/core/shelf.sol (hunk 14 above)
Comment (lucasvo) [reply src/core/shelf.sol]: Why not do that now?

File: src/core/shelf.sol (hunk 35 above)
Comment (lucasvo) [reply src/core/shelf.sol]: @manuelpolzhofer that was intentional. If we make it an auth method we have to add the admit functionality back and it doesn't give us a lot of benefits. I don't think we have to worry about the UI at the smart contract level :)

File: src/core/admin.sol [hunk 45]
```diff
@@ -72,22 +67,22 @@ contract Admin {
...
+        
+        pile.setRate(loan, rate);
         emit Whitelisted(loan);
         return loan;
     }
 
     function file(uint rate, uint speed) public auth {
-        debtRegister.file(rate, speed);
+        pile.file(rate, speed);
     }
```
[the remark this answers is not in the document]
ilinzweilin: pile.file sets the the speed for a rateGroup
pile.setRate assignes a loan to a certain rate

src/core/ceiling/creditline.sol
ilinzweilin: the repay function is part of the ceiling interface that this contract implements

src/core/admin.sol
```diff
@@ -98,7 +93,7 @@ contract Admin {
     function blacklist(uint loan) public auth {
         admit.update(loan, address(0), 0, 0);
         appraiser.file(loan, 0);
```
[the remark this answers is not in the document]
ilinzweilin: don't we want to continue keeping track of the values of the locked nfts?

src/core/pile.sol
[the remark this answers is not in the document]
ilinzweilin: Sorry, I was actually referring to the file function that I think might require changes.  Currently when we changing the speed of a rate group we just resetting chi to 1, which leads to issue when the group already has debt.

For this function (setRate) I think we dont even need to change anything.  We can keep the 0 rate group and the  `require(pie[loan]==0)` check should be enough. Why not being able to change the rate group of a loan if there is no debt on it
@manuelpolzhofer, @lucasvo 
What do you think?

src/core/pile.sol
ilinzweilin: agree. See comment above:

Currently when we changing the speed of a rate group we just resetting chi to 1, which leads to issue when the group already has debt.

ilinzweilin: and we also need to add a method that safely updates the group speed

src/core/shelf.sol
ilinzweilin: @manuelpolzhofer @lucasvo ceiling is calling pile.debt(loan) which will always return the most recent debt using the most recent chi -> pile.debt(loan) is calling compound internally without updating the state. I think the implementation would work as expected.... 

We need to call accrue, as pile.incDebt(loan, wad) requires the most recent chi to calculate the pie of the borrow amount... 

Can change the order though to avoid the compound call in pile.debt(loan) as chi will already be updated by the accrue call :) The reason why I put the pile.accrue call below ceiling.borrow call in first place is to avoid a redundant execution of the accrue function in case the ceiling.borrow check fails.

src/core/admit.sol
```diff
@@ -37,43 +42,46 @@ contract Admit {
...
 
     function depend (bytes32 what, address addr) public auth {
         if (what == "shelf") { shelf = ShelfLike(addr); }
+        if (what == "ceiling") { ceiling = CeilingLike(addr); }
         else revert();
     }
     
     // --- Admit ---
-    function admit (address registry, uint nft, uint principal, address usr) public auth returns (uint) {
+    function admit(address registry, uint nft, uint principal, address usr) public auth returns (uint) {
```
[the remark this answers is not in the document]
ilinzweilin: I think we will get rid of admit anyway

src/core/pile.sol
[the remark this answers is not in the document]
manuelpolzhofer: agree with Alina. Discussed one on one

src/core/admin.sol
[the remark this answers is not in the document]
manuelpolzhofer: discussed with Lucas

src/core/shelf.sol
manuelpolzhofer: Agree with Alina. `debt() function guarantees the latest debt

src/core/shelf.sol
manuelpolzhofer: @alina Lets give full access in the deployer

src/core/shelf.sol
```diff
@@ -124,7 +130,7 @@ contract Shelf is DSNote, TitleOwned {
     }
 
     function borrow(uint loan, uint wad) public owner(loan) {
-        require(nftLocked(loan), "nft-not-locked");
+        require(_nftLocked(loan), "nft-not-locked");
```
lucasvo: I think nftLocked should be a public method. It can be helpful for other contracts and there is no harm in view only public methods.

src/core/admin.sol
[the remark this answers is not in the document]
ilinzweilin: removed appraiser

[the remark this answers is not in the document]
manuelpolzhofer: top

src/core/shelf.sol
ilinzweilin: as discussed with @manuelpolzhofer moving to separate ticket

src/core/shelf.sol
```diff
@@ -117,13 +118,15 @@ contract Shelf is DSNote, TitleOwned {
         return loan;
     }
 
-    function close(uint loan) public owner(loan) {
+    function close(uint loan) public {
         require(pile.debt(loan) == 0, "outstanding-debt");
+        (address registry, uint tokenId) = token(loan);
+        require(title.ownerOf(loan) == msg.sender || NFTLike(registry).ownerOf(tokenId) == msg.sender, "not loan or nft owner");
```
manuelpolzhofer: okay, lets keep it for now

src/core/deployer.sol
```diff
@@ -70,11 +71,11 @@ contract ShelfFab {
...
-contract DeskFab {
-    // note: this is the mock Desk, which will interface with the lender/tranche side of Tinlake, and does not require auth for now.
-    function newDesk(address shelf_, address token_) public returns (Desk desk) {
-        desk = new Desk(shelf_, token_);
-        return desk;
+contract TrancheManagerFab {
+    // note: this is the mock TrancheManager, which will interface with the lender/tranche side of Tinlake, and does not require auth for now.
+    function newTrancheManager(address shelf_, address token_) public returns (TrancheManager trancheManager) {
+        trancheManager = new TrancheManager(shelf_, token_);
+        return trancheManager;
```
manuelpolzhofer: In the end, it is the tranche which actually would require the transaction.

We need to pass two addresses in the current setup to the shelf

PR #173, opened 2020-01-12
chore: tranche tests and assessor mock

charleenfei:


lucasvo:
Wouldn't it make sense to build this on top of Manuel's changes?

src/core/lender/test/mock/assessor.sol
```diff
@@ -1,14 +1,57 @@
...
+        uint poolValue = returnValues["pileDebt"];
+        if (tranche == junior) {
+            return calcJuniorAssetValue(poolValue, trancheReserve, seniorDebt());
+        }
+        return calcSeniorAssetValue(poolValue, trancheReserve, seniorDebt(), juniorReserve());
+    }
+
+    // Tranche.assets (Junior) = (Pool.value + Tranche.reserve - Senior.debt) > 0 && (Pool.value - Tranche.reserve - Senior.debt) || 0
+    function calcJuniorAssetValue(uint poolValue, uint trancheReserve, uint seniorDebt) internal returns (uint) {
+        int assetValue = int(poolValue + trancheReserve - seniorDebt);
```
lucasvo: I think if this is just supposed to be a mock, it shouldn't need to implement this logic?

lucasvo: maybe we should consider renaming `trancheReserve` to `juniorReserve`, wdyt?

lucasvo: @manuelpolzhofer we also have a todo to find a different name for `poolValue` as we don't call it a pool anywhere else.

src/core/lender/tranche/test/tranche.t.sol
```diff
@@ -0,0 +1,93 @@
...
+    Tranche tranche;
+    address tranche_;
+    SimpleToken token;
+    SimpleToken currency;
+
+    address self;
+
+    function setUp() public {
+        // Simple ERC20
+        token = new SimpleToken("TIN", "Tranche", "1", 0);
```
lucasvo: Can we tinlake-erc20 for this instead?

src/core/lender/test/mock/assessor.sol
```diff
@@ -1,14 +1,57 @@
...
+        return (assetValue > 0) ? uint(assetValue) : 0;
+    }
+    // Tranche.assets (Senior) = (Tranche.debt < (Pool.value + Junior.reserve)) && (Senior.debt + Tranche.reserve) || (Pool.value + Junior.reserve + Tranche.reserve)
+    function calcSeniorAssetValue(uint poolValue, uint trancheReserve, uint trancheDebt, uint juniorReserve) internal returns (uint) {
+        return ((poolValue + juniorReserve) >= trancheDebt) ? (trancheDebt + trancheReserve) : (poolValue + juniorReserve + trancheReserve);
+    }
+    function juniorReserve() internal returns (uint) {
+        return call("juniorReserve");
+    }
+    function seniorDebt() internal returns (uint) {
```
lucasvo: Is there actually a reason this needs to exist on the assessor at all? (I know I'm reviewing the mock - not the impl) but I'm curious!

src/core/lender/test/mock/assessor.sol
charleenfei: > I think if this is just supposed to be a mock, it shouldn't need to implement this logic?

yea you're right, I changed it in the next PR

charleenfei:
> Wouldn't it make sense to build this on top of Manuel's changes?

just opening this PR for review, I can rebase it when @manuelpolzhofer's PR is merged in

src/core/lender/test/mock/assessor.sol
manuelpolzhofer: we need it in the junior calculation for the case where no senior tranche exists
```javascript
  function seniorDebt() internal returns (uint) {
        return (senior != address(0x0)) ? SeniorTrancheLike(senior).debt() : 0;
    }
```

we could try to call `SeniorTrancheLike(0x00...000).debt()` and see if it fails. maybe it would return just
a 0 because the contract doesn't exists.

I prefer the ckeck, I think

src/core/lender/test/mock/assessor.sol
```diff
@@ -1,14 +1,57 @@
...
+
+
+    function setReturn(bytes32 name, uint returnValue) public {
+        returnValues[name] = returnValue;
+    }
+
+    function call(bytes32 name) internal returns (uint) {
+        calls[name]++;
+        return returnValues[name];
+    }
```
manuelpolzhofer: nice :) 

I like the pattern.

src/core/lender/tranche/test/tranche.t.sol
manuelpolzhofer: `SimpleToken` is only for testing. No `auth` on the mint method to make it easier

src/core/lender/tranche/test/tranche.t.sol
```diff
@@ -0,0 +1,93 @@
...
+        currency = new SimpleToken("CUR", "Currency", "1", 0);
+        tranche = new Tranche(address(token), address(currency));
+        tranche_ = address(tranche);
+
+        self = address(this);
+    }
+
+    function testBalance() public {
+        currency.mint(tranche_, 100);
```
manuelpolzhofer: let's try to always use numbers with `ether` to be closer to how it will be used.
```suggestion
        currency.mint(tranche_, 100 ether);
```

src/core/lender/test/mock/assessor.sol
charleenfei: > I think if this is just supposed to be a mock, it shouldn't need to implement this logic?

I think all the values are anyway passed down to the function that calls these functions, so rather than mocking the whole thing we can just use those values. wdyt? I prefixed now the method names so that it is clear they are internal.

src/core/lender/tranche/test/tranche.t.sol
```diff
@@ -0,0 +1,93 @@
...
+
+    function testBorrow() public {
+        currency.mint(tranche_, 100);
+        currency.approve(tranche_, uint(-1));
+        assertEq(currency.balanceOf(tranche_), 100);
+        tranche.borrow(self, 100);
+        assertEq(currency.balanceOf(tranche_), 0);
+        assertEq(currency.balanceOf(self), 100);
+    }
+}
```
ilinzweilin: I don't think that you need any of these currency.approve() & token.approve() functions

charleenfei: rm in testBorrow

src/core/lender/tranche/test/tranche.t.sol
```diff
@@ -0,0 +1,92 @@
...
+        currency.approve(tranche_, uint(-1));
+        token.approve(tranche_, uint(-1));
+        token.mint(self, 50 ether);
+        tranche.redeem(self, 100 ether, 50 ether);
+        assertEq(currency.balanceOf(self), 100 ether);
+        assertEq(token.balanceOf(self), 0);
+    }
+
+    function testRepay() public {
+        currency.mint(self, 100);
```
manuelpolzhofer:
```suggestion
        currency.mint(self, 100 ether);
```

PR #196, opened 2020-01-14
chore: remove OpenZeppelin directory in src/core

charleenfei:


.gitmodules
```diff
@@ -31,3 +31,6 @@
 [submodule "lib/tinlake-math"]
 	path = lib/tinlake-math
 	url = https://github.com/centrifuge/tinlake-math
+[submodule "lib/tinlake-simple-nft"]
+	path = lib/tinlake-simple-nft
+	url = https://github.com/centrifuge/tinlake-simple-nft
```
manuelpolzhofer: How much work would it be to use title here?

The only difference between title and simple-nft are the permissions for `mint`

PR #198, opened 2020-01-16
feat: added initial NAV to assessor

manuelpolzhofer:


src/core/lender/assessor.sol
```diff
@@ -71,6 +79,10 @@ contract Assessor is DSNote,DSMath {
     }
 
     function calcTokenPrice() public returns (uint) {
+        return mul(_calcTokenPrice(), initialNAV);
```
charleenfei: rmul?

PR #191, opened 2020-01-14
feat: added simple price pool contract

manuelpolzhofer:

PR #199, opened 2020-01-16
feat: assessor uses price pool

manuelpolzhofer:
please review: #198 (NAV) and #191 (pricePool) first


src/core/lender/assessor.sol
```diff
@@ -30,6 +30,10 @@ contract PileLike {
     function debt() public returns(uint);
 }
 
+contract PoolLike {
+    function totalValue() public returns(uint);
+}
+
 contract Assessor is DSNote,DSMath {
```
ilinzweilin: Why not tinlake-math?

PR #202, opened 2020-01-16
use tinlake-math in pile

ilinzweilin:


src/core/pile.sol
```diff
@@ -78,19 +79,19 @@ contract Pile is DSNote {
         uint rate = group[loan];
         uint chi = rates[rate].chi;
         if (now >= rates[rate].rho) {
-            (chi,) = _compound(rate);
+            (chi,) = compounding(rates[rate].chi, rates[rate].speed, rates[rate].rho, rates[rate].pie);
```
manuelpolzhofer:
```suggestion
            chi= updateChi(rates[rate].chi, rates[rate].speed, rates[rate].rho, rates[rate].pie);
```
We have a own method which only update chi and doesn't calculate the `delta`

src/core/pile.sol
```diff
@@ -78,19 +79,19 @@ contract Pile is DSNote {
...
+        return toAmount(chi, pie[loan]);
     }
 
     function rateDebt(uint rate) public view returns (uint) {
         uint chi = rates[rate].chi;
         uint pie = rates[rate].pie;
         
         if (now >= rates[rate].rho) {
-            (chi,) = _compound(rate);
+            (chi,) = compounding(rates[rate].chi, rates[rate].speed, rates[rate].rho, pie);
```
manuelpolzhofer: same here

PR #186, opened 2020-01-14
fix: correct distributor behaviour && distributor tests

manuelpolzhofer:
please review #177 first
- introduce mock base
- unit tests for distributor single tranche
- unit tests for distributor two tranches
- only repay if currencyAmount > 0

PR #204, opened 2020-01-16
Removing balance() method from tranche

lucasvo:


src/core/lender/distributor/base.sol
```diff
@@ -108,9 +115,10 @@ contract Distributor is DSNote, Math {
     function repayTranches(uint available) public auth {
         // repay senior always first
         if(address(senior) != address(0)) {
-            if(junior.balance() > 0 && senior.debt() > 0) {
+            uint juniorBalance = currency.balanceOf(address(junior));
```
lucasvo: Having the call twice here costs gas, we should try to avoid that.

src/core/lender/distributor/switchable.sol
```diff
@@ -16,18 +16,14 @@
 pragma solidity >=0.4.24;
 
 import "ds-note/note.sol";
-import "./base.sol";
+import { CurrencyLike, Distributor, TrancheLike } from "./base.sol";
```
ilinzweilin: do we need the CurrencyLike & TrancheLike imports here?

manuelpolzhofer:
All changes are applied here in the latest PR:
https://github.com/centrifuge/tinlake/pull/186/

I would like to close this PR.

src/core/lender/distributor/base.sol
manuelpolzhofer: I have a PR open which fixes it.

src/core/lender/distributor/switchable.sol
```diff
@@ -39,15 +35,9 @@ contract SwitchableDistributor is Distributor {
...
-    function depend(bytes32 what, address addr) public auth {
-        if (what == "currency") {
-            currency = CurrencyLike(currency);
-        }  else revert();
-    }
-
     function balance() public {
         if(borrowFromTranches) {
-            uint currencyAmount = add(senior.balance(), junior.balance());
+            uint currencyAmount = add(currency.balanceOf(address(senior)), currency.balanceOf(address(junior)));
```
manuelpolzhofer: ahh didn't see, that one. I would prefer if the tranche can tell how much currency is available to take.

It allows more flexibility in the tranche impl.

PR #210, opened 2020-01-17
feat: tinlake math and upgrade solidity version

manuelpolzhofer:


src/core/lender/tranche/test/senior_tranche.t.sol
```diff
@@ -84,63 +84,37 @@ contract SeniorTrancheTest is DSTest, DSMath {
...
-                    let zxRound := add(zx, half)
-                    if lt(zxRound, zx) { revert(0,0) }
-                    z := div(zxRound, base)
-                }
-            }
-            }
-        }
-    }
-
+    // TODO needs to be fixed with rounding error
```
manuelpolzhofer: I will handle the fix in a seperated PR

PR #219, opened 2020-01-17
fix: remove collector operator & add custom collector auth

ilinzweilin:


src/core/borrower/collect/test/collector.t.sol
```diff
@@ -82,6 +83,7 @@ contract CollectorTest is DSTest {
     }
 
     function testSeizeCollectAnyUser() public {
+        collector.relyCollector(address(this));
```
manuelpolzhofer: This is an edge case.

The test is `ward` and `investor` add the same time.

maybe a `UserCollector` with a `doCollect`?

PR #221, opened 2020-01-18
feat: system tests use new deployers and lender contracts

manuelpolzhofer:
**Deployer: Lender, Borrower**
- Deployer using auth pattern
- the user who created the deployer doesn't have access after deployment is done
- added permissions to deployer

**RootAdmin Contract**
- after borrower and lender are deployed only rootAdmin has access.
- root admin completes the deployment 
- root admin can give access to contracts

**Feat**
- pricePool Fab
- pricePool added to borrower deployer
- test_utils.sol
  - has methods to deploy all Tinlake contracts
  - handles the user and admin contract

**Fix**
- Assessor `calcTokenPrice` uses tranche parameter


src/core/deployer.sol
```diff
@@ -12,187 +11,68 @@
...
-        principal = new Principal();
-        principal.rely(msg.sender);
-        principal.deny(address(this));
-    }
-}
+        // Borrower  Depends
+        borrowerDeployer.collector().depend("distributor", distributor_);
+        // todo needs to be tranche after mock
+        borrowerDeployer.shelf().depend("lender", distributor_);
+        borrowerDeployer.collector().depend("distributor", distributor_);
```
ilinzweilin: dupl line 🙈

src/core/deployer.sol
```diff
@@ -12,187 +11,68 @@
...
-        title = titlefab.newTitle(name, symbol);
-        title.rely(god);
-    }
-
-    function deployLightSwitch() public {
-        lightswitch = lightswitchfab.newLightSwitch();
-        lightswitch.rely(god);
-    }
+        // distributor allowed to call
+        borrowerDeployer.shelf().rely(address(lenderDeployer.distributor()));
```
ilinzweilin: don't think that this is required. distributor just calls public methods from shelf

src/core/deployer.sol
```diff
@@ -12,187 +11,68 @@
...
 
-contract CollectorFab {
-    function newCollector(address distributor, address shelf, address pile, address threshold) public returns (Collector collector) {
-        collector = new Collector(distributor, shelf, pile, threshold);
-        collector.rely(msg.sender);
-        collector.deny(address(this));
-    }
-}
+    function wireDepends() public auth {
+        address distributor_ = address(lenderDeployer.distributor());
```
ilinzweilin: maybe we also need to make sure here that the addresses already exist

src/core/deployer.sol
```diff
@@ -12,187 +11,68 @@
...
-        shelf.rely(god);
     }
 
-    // note: this method will be refactored with the new lender side contracts, we will rely on God once more
-    //and the Pile should articulate that it depends on the distributor, not a generic "lender".
-    function deployDistributor(address currency_) public {
-        distributor = distributorFab.newDistributor(address(shelf), currency_);
-        shelf.depend("lender", address(distributor));
+    function denyBorrowAdmin(address usr) public auth {
+        borrowerDeployer.principal().deny(usr);
```
ilinzweilin: what about threshold and collector

src/core/deployer.sol
```diff
@@ -12,187 +11,68 @@
...
+        borrowerDeployer.pile().deny(usr);
     }
 
-    function deployPrincipal() public {
-        principal = principalFab.newPrincipal();
-        principal.rely(god);
+    function relyBorrowAdmin(address usr) public auth {
+        borrowerDeployer.principal().rely(usr);
+        borrowerDeployer.shelf().rely(usr);
+        borrowerDeployer.pile().rely(usr);
```
ilinzweilin: same

src/core/deployer.sol
```diff
@@ -12,187 +11,68 @@
...
-        address distributor_ = address(distributor);
-        address shelf_ = address(shelf);
-        address collector_ = address(collector);
+    // todo: currently only needed for testing.
+    // test wants to create a specific setup
+    function relyRootAdmin(address usr) public auth {
+        borrowerDeployer.title().rely(usr);
+        borrowerDeployer.principal().rely(usr);
+        borrowerDeployer.shelf().rely(usr);
+        borrowerDeployer.pile().rely(usr);
```
ilinzweilin: same

src/core/borrower/deployer.sol
```diff
@@ -0,0 +1,191 @@
...
+    function deployShelf(address currency_) public auth {
+        shelf = shelffab.newShelf(currency_, address(title), address(pile), address(principal));
+        shelf.rely(mainDeployer);
+    }
+
+    function deployPrincipal() public auth {
+        principal = principalFab.newPrincipal();
+        principal.rely(mainDeployer);
+    }
+
```
ilinzweilin: I think we are also missing the CreditLine fab here

src/core/deployer.sol
manuelpolzhofer: yes, I only added the once which are currently need to fix the tests.

I will add the rest

src/core/deployer.sol
manuelpolzhofer: yes. I will add more safety checks. It is only called once so it makes sense

src/core/borrower/deployer.sol
manuelpolzhofer: I think we should rename the `fabs`.

PrincipalFab and CreditLineFab should both have a method called `newCeiling`

In the deployer i would use an abstract contract called `CeilingFabLike`

src/core/lender/assessor.sol
```diff
@@ -78,13 +78,13 @@ contract Assessor is Math, DSNote, Auth {
         return _calcSeniorAssetValue(poolValue, trancheReserve, SeniorTrancheLike(tranche).debt(), _juniorReserve());
     }
 
-    function calcTokenPrice() public returns (uint) {
-        return mul(_calcTokenPrice(), tokenAmountForONE);
+    function calcTokenPrice(address tranche) public returns (uint) {
```
manuelpolzhofer: The method is called by the operator and not the tranche.

That's the reason why `msg.sender` is not working.

src/core/test/system/system.t.sol
```diff
@@ -15,194 +15,93 @@
```
manuelpolzhofer: The current systems are basically testing the borrower modules.

A method called `setupCurrencyOnLender` guarantees that the lender has currency for the tests

PR #222, opened 2020-01-18
fix: collector requires nftPrice > 0

manuelpolzhofer:
- added require for check
- added test
- changed to more descriptive naming

### DIFF
--- a/src/core/borrower/collect/collector.sol
+++ b/src/core/borrower/collect/collector.sol
@@ -54,7 +54,7 @@ contract Collector is DSNote, Auth {
     // --- Data ---
     RegistryLike threshold;
     struct Lot {
-        address usr;
+        address buyer;
         uint    nftPrice;
     }
     mapping (uint => Lot) public tags;
@@ -80,8 +80,8 @@ contract Collector is DSNote, Auth {
     }
 
     // --- Collector ---
-    function file(uint loan, address usr, uint nftPrice) public auth {
-        tags[loan] = Lot(usr, nftPrice);
+    function file(uint loan, address buyer, uint nftPrice) public auth {
+        tags[loan] = Lot(buyer, nftPrice);
     }
 
     function seize(uint loan) public {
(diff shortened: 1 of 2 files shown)


src/core/borrower/collect/collector.sol
```diff
@@ -53,11 +53,12 @@ contract Collector is DSNote, Auth {
...
-    struct Lot {
-        address usr;
-        uint    wad;
+
+    struct SellingPrice {
+        address buyer;
+        uint    nftPrice;
     }
-    mapping (uint => Lot) public tags;
+    mapping (uint => SellingPrice) public prices;
```
lucasvo: IMO is not a good refactor. Because it's not just prices but also users who may buy.

`SellingPrice` is grammatically not quite correct. It would have to be SalePrice or OfferPrice. A good compromise IMO could be:

Option: an option allows someone to execute a trade.

src/core/borrower/collect/collector.sol
```diff
@@ -90,11 +91,12 @@ contract Collector is DSNote, Auth {
...
     }
 
-    function collect(uint loan, address usr) public auth_collector {
-        require(usr == tags[loan].usr || tags[loan].usr == address(0));
+    function collect(uint loan, address buyer) public auth_collector {
+        require(buyer == prices[loan].buyer || prices[loan].buyer == address(0));
         (address registry, uint nft) = shelf.token(loan);
-        shelf.recover(loan, usr, tags[loan].wad);
-        NFTLike(registry).transferFrom(address(this), usr, nft);
+        require(prices[loan].nftPrice > 0, "no nft price defined");
```
lucasvo: didn't we say `contract/no-spaces-in-revert` in the styleguide?

src/core/borrower/collect/collector.sol
manuelpolzhofer: change it to option

src/core/borrower/collect/collector.sol
manuelpolzhofer: fixed

src/core/borrower/collect/collector.sol
```diff
@@ -53,11 +53,12 @@ contract Collector is DSNote, Auth {
 
     // --- Data ---
     RegistryLike threshold;
-    struct Lot {
-        address usr;
-        uint    wad;
+
+    struct Option {
```
lucasvo:
```suggestion
    // An `Option` allows the `buyer` to buy the NFT for the specified price from the collector by calling `seize` and then `recover`.
    struct Option {
```

PR #228, opened 2020-01-19
chore: two happy path system tests for lender side -- supply and redeem

charleenfei:


src/core/test/system/test_utils.sol
```diff
@@ -90,6 +90,31 @@ contract User is DSTest{
     }
 }
 
+contract WhitelistInvestor is DSTest{
```
manuelpolzhofer:
```suggestion
contract WhitelistInvestor is DSTest{
```
```suggestion
contract WhitelistedInvestor is DSTest{
```

src/core/lender/deployer.sol
```diff
@@ -93,11 +93,18 @@ contract DistributorLike {
...
     function deny(address usr) public;
 }
 
+contract WhitelistOperatorLike {
+    function rely(address usr) public;
+    function deny(address usr) public;
+    function relyInvestor(address usr) public;
+    function denyInvestor(address usr) public;
+}
+
```
manuelpolzhofer: I want to avoid that.

The idea is to have only an `OperatorFab`. 

You need to cast it in the test to a whitelist Operator

src/core/test/system/system.t.sol
```diff
@@ -321,4 +324,23 @@ contract SystemTest is TestUtils, DSTest {
         borrower.doBorrow(loan, 100);
         assertEq(currency.balanceOf(borrower_), 100);
     }
+
+    function testSupply() public {
```
manuelpolzhofer: lets make a own file for lender tests

src/core/test/system/test_utils.sol
manuelpolzhofer: I think we don't need to make it DSTest

src/core/test/system/test_utils.sol
```diff
@@ -90,6 +90,31 @@ contract User is DSTest{
...
+
+    constructor(address operator_, address currency_, address token_) public {
+        operator = WhitelistOperator(operator_);
+        currency = ERC20Like(currency_);
+        token = ERC20Like(token_);
+    }
+
+    function doSupply(uint amount) public {
+        address tranche_ = address(operator.tranche());
+        currency.approve(tranche_, uint(-1));
```
manuelpolzhofer:
```suggestion
        currency.approve(tranche_, amount);
```

src/core/test/system/test_utils.sol
```diff
@@ -90,6 +90,31 @@ contract User is DSTest{
...
+        operator.supply(amount);
+    }
+
+    function doRedeem(uint amount) public {
+        address tranche_ = address(operator.tranche());
+        currency.approve(tranche_, uint(-1));
```
manuelpolzhofer:
```suggestion

```
not needed

src/core/test/system/test_utils.sol
```diff
@@ -90,6 +90,31 @@ contract User is DSTest{
...
+        token.approve(tranche_, uint(-1));
```
manuelpolzhofer:
```suggestion
        token.approve(tranche_, amount);
```

src/core/test/system/system.t.sol
```diff
@@ -140,6 +138,11 @@ contract SystemTest is TestUtils, DSTest {
         return currency.balanceOf(address(lenderDeployer.distributor()));
     }
 
+    function supply(uint supply, uint amount) public {
```
manuelpolzhofer:
```suggestion
    function supply(uint balance, uint amount) public {
```

The naming is a bit confusing

src/core/test/system/system.t.sol
```diff
@@ -321,4 +324,23 @@ contract SystemTest is TestUtils, DSTest {
         borrower.doBorrow(loan, 100);
         assertEq(currency.balanceOf(borrower_), 100);
     }
+
+    function testSupply() public {
+        supply(100 ether, 10 ether);
```
manuelpolzhofer: I think we should use variable and not constant numbers if we have more than one parameter.

```javascript
uint investorBalance;
uint supplyAmount;
supply(investorBalance, supplyAmount);
```

src/core/test/system/system.t.sol
```diff
@@ -321,4 +324,23 @@ contract SystemTest is TestUtils, DSTest {
...
+    function testRedeem() public {
+        supply(100 ether, 10 ether);
+        assertEq(currency.balanceOf(juniorInvestor_), 90 ether);
+        assertEq(lenderDeployer.juniorERC20().balanceOf(juniorInvestor_), 10 ether);
+        // additional liquidity changes the token price
+        // TODO: add code to move funds from shelf to tranche so we can enact the actual repayment scenario
+        currency.mint(address(lenderDeployer.junior()), 100 ether);
+        juniorInvestor.doRedeem(10 ether);
+        assertEq(lenderDeployer.juniorERC20().balanceOf(juniorInvestor_), 0);
+        assertEq(currency.balanceOf(juniorInvestor_), 190 ether);
```
manuelpolzhofer: Shouldn't it be `200 ether`

I think the test is not working because you need to change the flag in the distributor to repay.

src/core/test/system/system.t.sol
```diff
@@ -43,18 +44,15 @@ contract SystemTest is TestUtils, DSTest {
...
 
-        // todo replace with investor contract
-        rootAdmin.relyLenderAdmin(address(this));
-
-        // give invest rights to test
-        WhitelistOperator juniorOperator = WhitelistOperator(address(lenderDeployer.juniorOperator()));
-        juniorOperator.relyInvestor(address(this));
-
+        // junior investor
+        rootAdmin.relyLenderAdmin(juniorInvestor_);
```
manuelpolzhofer: The investor should be an admin.  The investor should be only allowed to call the tranche.

src/core/test/system/root_admin.sol
```diff
@@ -29,6 +29,7 @@ contract TestRootAdmin is RootAdmin {
     // Needed for System Tests
     function relyLenderAdmin(address usr) public auth {
         lenderDeployer.juniorOperator().rely(usr);
+        lenderDeployer.juniorOperator().relyInvestor(usr);
```
manuelpolzhofer: I think an admin shouldn't be an investor by default

src/core/test/system/system.t.sol
```diff
@@ -43,18 +44,15 @@ contract SystemTest is TestUtils, DSTest {
...
         // give admin access rights to contract
         // root only for this test setup
         rootAdmin.relyBorrowAdmin(admin_);
 
-        // todo replace with investor contract
-        rootAdmin.relyLenderAdmin(address(this));
-
-        // give invest rights to test
-        WhitelistOperator juniorOperator = WhitelistOperator(address(lenderDeployer.juniorOperator()));
-        juniorOperator.relyInvestor(address(this));
```
manuelpolzhofer: why did you remove it?

src/core/test/system/lender/redeem.t.sol
```diff
@@ -0,0 +1,36 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+pragma solidity >=0.5.12;
+
+import "../system.t.sol";
+import {SwitchableDistributor} from "../../../lender/distributor/switchable.sol";
+
+contract RedeemTest is SystemTest {
```
manuelpolzhofer:
```suggestion
contract RedeemTest is TestUtils {
```

If we use system tests we execute all system test twice.

But maybe it is a good pattern to move a default `setUp` method into TestUtils

src/core/test/system/system.t.sol
```diff
@@ -101,11 +103,16 @@ contract SystemTest is TestUtils, DSTest {
...
+        // borrow transaction
         borrower.doBorrow(loan, principal);
         checkAfterBorrow(tokenId, principal);
     }
 
+    function supply(uint balance, uint amount) public {
+        currency.mint(juniorInvestor_, balance);
+        juniorInvestor.doSupply(amount);
+    }
+
```
manuelpolzhofer: I would move it to testUtils

src/core/lender/distributor/base.sol
```diff
@@ -150,11 +151,13 @@ contract BaseDistributor is Math, DSNote, Auth {
 
         // repay senior always first
         if(address(senior) != address(0)) {
+            currency.approve(address(senior), available);
```
manuelpolzhofer: maybe lets move it to `_repay`

src/core/test/system/lender/redeem.t.sol
charleenfei: after conversation with @ilinzweilin and @manuelpolzhofer we have decided to move this issue to another PR

src/core/test/system/lender/redeem.t.sol
```diff
@@ -0,0 +1,36 @@
...
+        uint supplyAmount = 10 ether;
+        supply(investorBalance, supplyAmount);
+        assertEq(currency.balanceOf(juniorInvestor_), 90 ether);
+        assertEq(lenderDeployer.juniorERC20().balanceOf(juniorInvestor_), 10 ether);
+
+        SwitchableDistributor switchable = SwitchableDistributor(address(lenderDeployer.distributor()));
+        switchable.file("borrowFromTranches", false);
+        juniorInvestor.doRedeem(10 ether);
+        assertEq(lenderDeployer.juniorERC20().balanceOf(juniorInvestor_), 0);
+        assertEq(currency.balanceOf(juniorInvestor_), 100 ether);
```
ilinzweilin: same as in the comment below, would be helpful to use the vars directly in the assertions

src/core/test/system/lender/redeem.t.sol
ilinzweilin: yes, I would even create a clean base SystemTest contract that inherits from testUtils but also has some additional methods like setUp, assertPrecondition, assertPostcondition... and let the tests inherit from it.

src/core/test/system/users/whitelisted_investor.sol
```diff
@@ -0,0 +1,42 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+pragma solidity >=0.5.12;
+
+import "ds-test/test.sol";
+import "../../../lender/deployer.sol";
+import "../test_utils.sol";
```
ilinzweilin: maybe you can also add the interfaces you need to the interfaces file in the test folder and import it, then you can remove the test_utils & deployer imports.

src/core/test/system/test_utils.sol
```diff
@@ -72,11 +71,8 @@ contract User is DSTest{
     }
 
     function doRepay(uint loan, uint wad, address usr) public {
-        emit log_named_uint("loan", wad);
```
ilinzweilin: thank you 🙈

src/core/test/system/users/whitelisted_investor.sol
charleenfei: good idea, I will add it in the TestUtils Pr

PR #233, opened 2020-01-20
feat: currency routing through distributor

manuelpolzhofer:
please review #230 first

-  erc20 currency routing through distributor
- added new mock for tokenContract
- adjusted tests


src/core/borrower/shelf.sol
```diff
@@ -82,7 +82,12 @@ contract Shelf is DSNote, Auth, TitleOwned, Math {
     }
 
     function depend (bytes32 what, address addr) public auth {
-        if (what == "lender") { lender = addr; }
+        if (what == "lender") {
+            lender = addr;
+            // todo review alternatives
+            tkn.approve(lender, uint(-1));
```
manuelpolzhofer: Let's discuss all options here. I don't like the current solution.

First of all, we need a `uint(-1)`.  The distributor should have the right to take the money always back not depending on if the currency has been borrowed or not.

I would like to move the approve to the constructor but it would add a dependency on the deployers.

As an alternative would add a flag and the approval can be only set once, which will be from the deployer

manuelpolzhofer: 1. keep it like that. (i just added it to ilustrate)
2. add a only once flag here. (my prefered)
3.  move it to the constructor.

PR #236, opened 2020-01-20
add borrower tests for issue & close

ilinzweilin:
- added additional check for shelf.close()  `require(!nftLocked(loan))`
- moved test user contracts out of test-utils file into sep folder 
- renamed TestUtils contract to TestSetup
- added tests for borrower: issue & close 
- added shared interface file for test contracts. We can add some naming conventions to it.

There are some todos for the close tests:
- will be resolved as soon as refactoring for borrow tests is done


src/core/borrower/shelf.sol
```diff
@@ -181,7 +179,7 @@ contract Shelf is DSNote, Auth, TitleOwned, Math {
     }
 
     // --- NFT actions ---
-    function lock(uint loan, address usr) public owner(loan) {
+    function lock(uint loan) public owner(loan) {
```
manuelpolzhofer: lets say A owns the NFT and the loan. (and he/she added the approve call)

A sends NFT to B. If B would give an approve.

A could still lock the NFT with this method. Correct?

If B would create a new loan with the NFT. Could A still call this method? Would the registry entry still exist?

src/core/test/system/borrower/close.t.sol
```diff
@@ -0,0 +1,90 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+pragma solidity >=0.5.12;
+
+import "../system.sol";
+
+contract CloseTest is SystemTest {
```
manuelpolzhofer: I would prefer renaming the `setUp` method in SystemTest and adding always a `setUp` method in a test.

Which might be just a one liner calling the `defaultSetUp` from System test

src/core/test/system/borrower/close.t.sol
```diff
@@ -0,0 +1,90 @@
...
+    
+    function closeLoan(uint loanId, uint tokenId, bytes32 lookupId) public {
+        assertPreCondition(loanId, tokenId, lookupId);
+        borrower.close(loanId);
+        assertPostCondition(loanId, tokenId, lookupId);
+    }
```
manuelpolzhofer: 👍  good pattern

src/core/test/system/users/borrower.sol
```diff
@@ -0,0 +1,72 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+pragma solidity >=0.5.12;
+
+import "ds-test/test.sol";
+import { Title } from "tinlake-title/title.sol";
+import "../interfaces.sol";
+
+contract Borrower is DSTest {
```
manuelpolzhofer:
```suggestion
contract Borrower {
```

If we don't use some log calls it is not required to add it

src/core/borrower/shelf.sol
manuelpolzhofer: Let's maybe to avoid all the edge case always require the following

```javascript
require(msg.sender == title.ownerOf(loan) == NFTLike(shelf[loan].registry).ownerOf(shelf[loan].tokenId))
```

src/core/test/system/users/admin.sol
```diff
@@ -0,0 +1,61 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+pragma solidity >=0.5.12;
+
+import "ds-test/test.sol";
+import { Title } from "tinlake-title/title.sol";
+import "../interfaces.sol";
+
+contract AdminUser is DSTest{
```
charleenfei: DSTest maybe not needed here

src/core/borrower/shelf.sol
ilinzweilin: I think it is already given, because the msg.sender has to be the nftOwner and the loanOwner for lock to work.
NFTLike(shelf[loan].registry).transferFrom(msg.sender, address(this), shelf[loan].tokenId); -> makes sure user is nftOwner
modifier: owner(loan) -> makes sure user is loanOwner

src/core/test/system/users/borrower.sol
ilinzweilin: will remove after finishing developing all the tests.

src/core/test/system/borrower/close.t.sol
ilinzweilin: will rename in next PR

PR #246, opened 2020-01-21
fix: reroute currency transfers to distributor, rather than directly from tranche to tranche

charleenfei:
closes #233 and #234


src/core/lender/distributor/base.sol
```diff
@@ -151,6 +154,7 @@ contract BaseDistributor is Math, DSNote, Auth {
 
         // repay senior always first
         if(address(senior) != address(0)) {
+            currency.approve(address(senior), available);
```
manuelpolzhofer:
```suggestion

```
Not required the approve is in the `_repay` method.

src/core/lender/distributor/base.sol
```diff
@@ -80,8 +80,11 @@ contract BaseDistributor is Math, DSNote, Auth {
             if (amount > juniorBalance) {
                 amount = juniorBalance;
             }
-            // move junior reserve to senior
-            senior.repay(address(junior), amount);
+            // move junior reserve to distributor -> senior
+            // using the borrow method due to allowance permissions
+            junior.borrow(address(this), amount);
```
manuelpolzhofer: 🥇

PR #247, opened 2020-01-22
add borrower tests - borrow

ilinzweilin:
- added test cases for borrow
- swapped address(this) with random user


src/core/test/system/borrower/borrow.t.sol
```diff
@@ -0,0 +1,151 @@
...
+        borrower.lock(loanId);
+    }
+
+    function assertPreCondition(uint loanId, uint tokenId, uint amount) public {
+        // assert: borrower loanOwner
+        assertEq(title.ownerOf(loanId), borrower_);
+        // assert: shelf nftOwner
+        assertEq(collateralNFT.ownerOf(tokenId), address(shelf));
+        // assert: borrowAmount <= ceiling
+        assert(amount <= ceiling.ceiling(loanId));
```
manuelpolzhofer: It should be the following
```suggestion
        assert(amount < ceiling.ceiling(loanId));
```
 because the require is  `require(values[loan] >= amount);` or we change the require in the ceilings

src/core/test/system/borrower/close.t.sol
```diff
@@ -23,12 +23,17 @@ contract CloseTest is SystemTest {
...
         
+    Borrower randomUser;
+    address randomUser_;
+
     function setUp() public {
         baseSetup();
-        borrower = new Borrower(address(borrowerDeployer.shelf()), address(lenderDeployer.distributor()), currency_, address(borrowerDeployer.pile()));
+        borrower = new Borrower(address(shelf), address(distributor), currency_, address(pile));
         borrower_ = address(borrower);
+        randomUser = new Borrower(address(shelf), address(distributor), currency_, address(pile));
```
charleenfei: 👍

src/core/test/system/borrower/borrow.t.sol
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    function testPartialBorrow() public {
+        uint ceiling = 200 ether;
+        // borrow amount smaller then ceiling
+        uint amount = ceiling / 2;
+        (uint tokenId, ) = issueNFT(borrower_);
+        uint loanId = borrower.issue(collateralNFT_, tokenId);
+        lockNFT(loanId);
+        admin.setCeiling(loanId, amount * 2);
```
manuelpolzhofer:
```suggestion
        admin.setCeiling(loanId, ceiling);
```

manuelpolzhofer: maybe easier to read

src/core/test/system/borrower/borrow.t.sol
```diff
@@ -0,0 +1,151 @@
...
+        uint loanId = randomUser.issue(collateralNFT_, tokenId);
+        // lock nft for random user
+        randomUser.lock(loanId); 
+        // admin sets loan ceiling
+        admin.setCeiling(loanId, ceiling);
+        // borrower tries to borrow against loan
+        borrow(loanId, tokenId, amount);
+    }
+
+    function testFailBorrowAmountTooHeigh() public {
```
manuelpolzhofer:
```suggestion
    function testFailBorrowAmountTooHigh() public {
```

src/core/test/system/users/borrower.sol
```diff
@@ -45,6 +45,10 @@ contract Borrower is DSTest {
     }
 
     function borrow(uint loan, uint amount) public {
+        shelf.borrow(loan, amount);
```
charleenfei: maybe  this function wrapper is not necessary

src/core/test/system/borrower/borrow.t.sol
manuelpolzhofer: I think we should change it in the principal and creditline to:

 `require(add(pile.debt(loan), amount) <= values[loan]);`

ilinzweilin: we discussed with @lucasvo and agreed to consider debt just for the creditLine, not for principal

PR #251, opened 2020-01-22
add borrower tests - balance

ilinzweilin:
- fix balanceRequest in shelf
- add balance tests for switchable distributor
- use safeMath in tests
- rename whitelistedInvestor to investor


src/core/test/system/borrower/balance.t.sol
```diff
@@ -0,0 +1,131 @@
...
+        uint giveAmount = currency.balanceOf(address(junior));
+        borrower.balance();
+        assertPostConditionGive(giveAmount, initialCurrencyBalance);
+    }
+
+    function assertPreConditionTake(uint takeAmount) public {
+        // assert: borrowFromTranches is active
+        assert(distributor.borrowFromTranches());
+        // assert: junior has enough capital
+        assert(currency.balanceOf(address(junior)) == takeAmount);
```
charleenfei: :  >=

src/core/test/system/borrower/balance.t.sol
```diff
@@ -0,0 +1,131 @@
...
+        assertEq(currency.balanceOf(address(junior)), 0);
+        // assert: junior received funds
+        assertEq(currency.balanceOf(address(shelf)), add(initialCurrencyBalance, takeAmount));
+    }
+
+    function assertPreConditionGive(uint giveAmount) public {
+        // assert: borrowFromTranches is inactive
+        assert(!distributor.borrowFromTranches());
+        // assert: shelf has funds
+        assert(currency.balanceOf(address(shelf)) == giveAmount);
```
charleenfei: : >=

src/core/test/system/borrower/balance.t.sol
charleenfei: to test for surplus cases, only enough is transferred to cover debts

src/core/test/system/borrower/balance.t.sol
```diff
@@ -0,0 +1,131 @@
...
+        // assert: junior has enough capital
+        assert(currency.balanceOf(address(junior)) == takeAmount);
+    }
+
+    function assertPostConditionTake(uint takeAmount, uint initialCurrencyBalance) public {
+        // assert: takeAmount > 0
+        assert(takeAmount > 0);
+        // assert: all funds transferred from tranche reserve
+        assertEq(currency.balanceOf(address(junior)), 0);
+        // assert: junior received funds
```
charleenfei: shelf received funds

src/core/test/system/borrower/balance.t.sol
```diff
@@ -0,0 +1,131 @@
...
+        uint takeAmount = 100 ether;
+        // supply junior tranche with funds
+        supplyFunds(takeAmount, address(junior));
+        distributor.file("borrowFromTranches", false);
+        balanceTake();
+    }
+
+    function testBalanceGive() public {
+        uint giveAmount = 100 ether;
+        // supply shelf tranche with funds
```
charleenfei: supply tranche or shelf?

src/core/test/system/borrower/balance.t.sol
```diff
@@ -0,0 +1,131 @@
...
+        supplyFunds(giveAmount, address(junior));
+        // deactivate borrow from tranches
+        distributor.file("borrowFromTranches", false);
+        // take money from shelf and transfer into shelf
```
charleenfei: shouldn't this test be testing repay? then the shelf should have money to transfer to junior

src/core/test/system/borrower/balance.t.sol
```diff
@@ -0,0 +1,131 @@
...
+    function testFailBalanceGiveShelfHasNoFunds() public {
+        distributor.file("borrowFromTranches", false);
+        // do not supply shelf with funds
+        balanceGive();
+    }
+
+    function testFailBalanceGiveInactive() public {
+         uint giveAmount = 100 ether;
+        // supply shelf tranche with funds
+        supplyFunds(giveAmount, address(junior));
```
charleenfei: address(shelf), I think we are testing repay here?

src/core/test/system/borrower/balance.t.sol
```diff
@@ -0,0 +1,131 @@
...
+        // do not deactivate borrow
+        // take money from shelf and transfer into shelf
```
charleenfei: transfer to junior?

src/core/test/system/borrower/borrow.t.sol
```diff
@@ -137,10 +137,10 @@ contract BorrowTest is SystemTest {
         borrow(loanId, tokenId, amount);
     }
 
-    function testFailBorrowAmountTooHeigh() public {
```
charleenfei: 🎉

src/core/test/system/borrower/balance.t.sol
ilinzweilin: no, I added a comment above the supply Funds function. repay is calling internally balance(). So we can not use it here. Same applies for supply

src/core/test/system/borrower/balance.t.sol
ilinzweilin: in case of switchable distributor we always take and give the full available amount

src/core/test/system/borrower/balance.t.sol
ilinzweilin: tranche :)

src/core/test/system/borrower/balance.t.sol
charleenfei: I think we should be explicit about that it is testing the switchable distributor setup

src/core/test/system/borrower/balance.t.sol
ilinzweilin: good catch. I supplied the opposite side with funds

src/core/borrower/shelf.sol
```diff
@@ -129,10 +129,12 @@ contract Shelf is DSNote, Auth, TitleOwned, Math {
     
     // --- Currency actions ---
     function balanceRequest() public returns (bool, uint) {
-        if (balance > 0) {
-            return (true, balance);
+        uint currencyBalance = currency.balanceOf(address(this));
```
manuelpolzhofer: 💯

src/core/lender/deployer.sol
```diff
@@ -93,11 +93,13 @@ contract SwitchableDistributorFab {
 // todo add BaseDistributorFab
 
 contract DistributorLike {
+    bool public borrowFromTranches;
```
manuelpolzhofer: This should not be here. 

If it is the abstract contract

src/core/test/system/borrower/balance.t.sol
```diff
@@ -0,0 +1,132 @@
...
+        assert(distributor.borrowFromTranches());
+        // assert: tranche reserve has enough funds
+        assert(currency.balanceOf(address(junior)) == takeAmount);
+    }
+
+    function assertPostConditionTake(uint takeAmount, uint initialShelfBalance) public {
+        // assert: takeAmount > 0
+        assert(takeAmount > 0);
+        // assert: all funds transferred from tranche reserve
+        assertEq(currency.balanceOf(address(junior)), 0);
```
manuelpolzhofer: Not in every case but If you only test cases where it is true. 

It should be okay

src/core/test/system/borrower/balance.t.sol
```diff
@@ -0,0 +1,132 @@
...
+        // supply shelf with funds
+        supplyFunds(giveAmount, address(shelf));
+        // do not deactivate borrow
+        balanceGive();
+    }
+
+    // Helper to supply shelf or tranches with currency without using supply or repay, since these functions are usign balance internally.
+    function supplyFunds(uint amount, address addr) public {
+        currency.mint(address(this), amount);
+        currency.transferFrom(address(this), address(addr), amount);
```
manuelpolzhofer: I think this can be done in one line.

` currency.mint(address(addr), amount);`

src/core/test/system/borrower/balance.t.sol
ilinzweilin: in the switchable it is always the case :)

PR #257, opened 2020-01-23
add borrower tests repay & withdraw

ilinzweilin:
- feature: add balance() to repay & withdraw in shelf
- add system tests for withdraw & repay

PR #259, opened 2020-01-23
add borrower tests - unlock nft

ilinzweilin:
- please review #257 first. This PR is based on it


src/core/test/system/borrower/unlock.t.sol
```diff
@@ -0,0 +1,162 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+pragma solidity >=0.5.12;
+
+import "../system.sol";
+import "../users/borrower.sol";
+
+contract UnlockTest is SystemTest {
```
manuelpolzhofer: I think we should test the case, where the NFT is locked in the collector

ilinzweilin: yes, there is a todo in the cases for it. Coming in the PR with collector tests :)

PR #260, opened 2020-01-23
feat: assessor calculates interest delta for senior tranche

manuelpolzhofer:


src/core/lender/assessor/assessor.sol
```diff
@@ -0,0 +1,30 @@
...
+pragma solidity >=0.5.12;
+
+import "./base.sol";
+
+contract Assessor is BaseAssessor {
+
+    // accrueTrancheInterest can implement different interest models
+    function accrueTrancheInterest(address tranche_) public returns (uint) {
+        SeniorTrancheLike tranche = SeniorTrancheLike(tranche_);
+        uint debt = tranche.debt();
```
ilinzweilin: `assessor.accrueTrancheInterest` calls`tranche.debt()` which calls `drip()` which calls `assessor.accrueTrancheInterest()` 🙈

manuelpolzhofer: True. 👍  It should be `uint debt = `tranche.borrowed()+tranche.interest()`.

src/core/lender/assessor/default.sol
```diff
@@ -0,0 +1,31 @@
...
+
+contract DefaultAssessor is BaseAssessor, Interest {
+
+    // accrueTrancheInterest can implement different interest models
+    function accrueTrancheInterest(address tranche_) public returns (uint) {
+        SeniorTrancheLike tranche = SeniorTrancheLike(tranche_);
+        uint debt = tranche.borrowed() + tranche.interest();
+        // move to tinlake-math
+        // interest is calculated based on tranche debt
+        return safeSub(rmul(rpow(tranche.ratePerSecond(), now - tranche.lastUpdated(), ONE), debt), debt);
```
charleenfei: if we update the naming in tinlake-math, we can directly use `safeSub(updateChi(ONE, ratePerSecond, lastUpdated), debt)`

manuelpolzhofer: yes. we need to make the method more general.

It is not updateChi anymore.

charleenfei: yes, exactly

PR #264, opened 2020-01-24
feat: full investment assessor

manuelpolzhofer:
full investment assessor implements the behaviour of switchableDistributor interest calc


src/core/lender/assessor/full_investment.sol
```diff
@@ -0,0 +1,33 @@
...
+    // accrueTrancheInterest can implement different interest models
+    function accrueTrancheInterest(address tranche_) public returns (uint) {
+        SeniorTrancheLike tranche = SeniorTrancheLike(tranche_);
+        if(tranche_ == junior) {
+            return 0;
+
+        }
+        uint interestBearingAmount = safeAdd(safeAdd(tranche.borrowed(), tranche.interest()), tranche.balance());
+
+        return safeSub(rmul(rpow(tranche.ratePerSecond(), now - tranche.lastUpdated(), ONE), interestBearingAmount), interestBearingAmount);
```
charleenfei: I think you can use the tinlake-math method here

PR #265, opened 2020-01-24
feat: senior tranche setup

charleenfei:
- also updates naming of BaseDistributor -> DefaultDistributor


src/core/test/system/lender/base/redeem_two_tranche.sol
```diff
@@ -0,0 +1,66 @@
...
+    Investor juniorInvestor;
+    address  juniorInvestor_;
+
+    Investor seniorInvestor;
+    address  seniorInvestor_;
+
+    function setUp() public {
+        bytes32 operator_ = "whitelist";
+        bytes32 distributor_ = "default";
+        baseSetup(operator_, distributor_, true);
```
manuelpolzhofer: Is the baseSetup setting up one or two tranches by default?

src/core/test/system/lender/base/redeem_single_tranche.t.sol
```diff
@@ -55,33 +54,20 @@ contract RedeemTest is SystemTest {
...
-        // assert: no more tokens left for junior investor
-        assertEq(juniorERC20.balanceOf(juniorInvestor_), 0);
         // assert: back to original balance
         assertEq(currency.balanceOf(juniorInvestor_), investorBalance);
         // assert: shelf is balanced, excess has either been transferred to tranches or needed money transferred from distributor -> shelf
         assertEq(shelf.balance() - currency.balanceOf(address(shelf)), 0);
     }
 
-    function testFailNotEnoughToken() public {
+    function testFailNotEnoughMoney() public {
```
manuelpolzhofer: Lets stick to `currency` and `token` naming.s

src/core/test/system/lender/base/redeem_two_tranche.sol
manuelpolzhofer: Okay, now I understand. 
It would be easier to read
```javascript
bool deploySeniorTranche = true;
baseSetup(operator_, distributor_,  deploySeniorTranche);
```

src/core/test/system/setup.sol
```diff
@@ -133,11 +139,40 @@ contract TestSetup {
...
+        } else if (operator_ == "allowance") {
+            sOperatorfab_ = address(new AllowanceOperatorFab());
+        }
+
+        tranchefab_ = address(new SeniorTrancheFab());
+
+        lenderDeployer.depend("senior_tranche_fab", tranchefab_);
+        lenderDeployer.depend("senior_operator_fab", sOperatorfab_);
+
+        lenderDeployer.deploySeniorTranche("SUN", "Senior Tranche Token");
```
manuelpolzhofer: like the name :)

src/core/test/system/setup.sol
```diff
@@ -133,11 +139,40 @@ contract TestSetup {
...
+
+    function deploySenior(bytes32 operator_) public {
+        address sOperatorfab_;
+        address tranchefab_;
+
+        if (operator_ == "whitelist") {
+            sOperatorfab_ = address(new WhitelistOperatorFab());
+        } else if (operator_ == "allowance") {
+            sOperatorfab_ = address(new AllowanceOperatorFab());
+        }
```
manuelpolzhofer: Currently, we always deploy both tranches with the same operator.

We could just pass the operatorFab instead of doing the if else again.

PR #270, opened 2020-01-25
collect refactor & tests

ilinzweilin:
- features: rest loan balance on recover & close
- allow recoveryAmount bigger than outstanding loan debt
- add system tests for collector
- added keeper user
- add missing system tests for unlock & close
- remove DSTest dependency rom test users


src/core/borrower/shelf.sol
```diff
@@ -168,9 +169,15 @@ contract Shelf is DSNote, Auth, TitleOwned, Math {
...
     function recover(uint loan, address usr, uint currencyAmount) public auth {
-        _repay(loan, usr, currencyAmount);
-        uint loss = pile.debt(loan);
-        pile.decDebt(loan, loss);
+        pile.accrue(loan);
+        uint loanDebt = pile.debt(loan);
+        
+        currency.transferFrom(usr, address(this), currencyAmount);
+       
+        ceiling.repay(loan, currencyAmount);
```
manuelpolzhofer: Not quite sure here but i think it makes sense

ilinzweilin: I also thought about this line. Currently nothing is happening on ceiling.repay. But maybe we introduce some logic later that also is relevant for recovery so I left it in here

src/core/borrower/shelf.sol
```diff
@@ -203,4 +210,12 @@ contract Shelf is DSNote, Auth, TitleOwned, Math {
     function claim(uint loan, address usr) public auth {
         NFTLike(shelf[loan].registry).transferFrom(address(this), usr, shelf[loan].tokenId);
     }
+
+    function resetLoanBalance(uint loan) internal {
```
lucasvo: I think this method can be rewritten to be simpler:
```
function resetLoanBalance(uint loan) internal {
if (loanBalance == 0) return;
balance = safeSub(balance, balances[loan]);
balance[loan=0];
}
```

PR #272, opened 2020-01-25
test: interest tests + public drip senior method

manuelpolzhofer:
please review: #271


src/core/test/system/lender/assessor/default_interest.sol
```diff
@@ -0,0 +1,166 @@
...
+        hevm = Hevm(0x7109709ECfa91a80626fF3989D68f67F5b1DD12D);
+        hevm.warp(1234567);
+
+        assessor = TAssessorLike(address(lenderDeployer.assessor()));
+
+        createTestUsers(deploySeniorTranche);
+    }
+
+    function supplySenior(uint amount) public {
+        currency.mint(seniorInvestor_, amount);
```
charleenfei: i think these two functions could also be moved to the utils with doRedeem

src/core/test/system/lender/assessor/full_investment_interest.sol
```diff
@@ -0,0 +1,143 @@
...
+
+        hevm = Hevm(0x7109709ECfa91a80626fF3989D68f67F5b1DD12D);
+        hevm.warp(1234567);
+
+        assessor = TAssessorLike(address(lenderDeployer.assessor()));
+
+        createTestUsers(deploySeniorTranche);
+    }
+
+    function supplySenior(uint amount) public {
```
charleenfei: see above comment about util functions

src/core/test/system/lender/assessor/default_interest.sol
charleenfei: (supplyJunior below)

PR #273, opened 2020-01-26
feat: security impr tokenAmountForONE only set in constructor

manuelpolzhofer:
- no file method for variables which should only be set once
- any recommendations for a better variable name `tokenAmountForONE`


src/core/lender/assessor/base.sol
```diff
@@ -89,6 +91,7 @@ contract BaseAssessor is Math, Auth {
     }
 
     function calcTokenPrice(address tranche) public returns (uint) {
+        require(tranche  == junior || tranche == senior);
```
manuelpolzhofer: I think the assessor should only give you numbers for senior and junior.

Theoretically, somebody could implement the interface in a contract and we would return a number

lucasvo: yes

src/core/lender/assessor/default.sol
```diff
@@ -20,6 +20,8 @@ import "tinlake-math/interest.sol";
 
 contract DefaultAssessor is BaseAssessor, Interest {
 
+    constructor(uint tokenAmountForONE) BaseAssessor(tokenAmountForONE) public {}
+
     // accrueTrancheInterest can implement different interest models
     function accrueTrancheInterest(address tranche_) public returns (uint) {
```
charleenfei: should we also do the same address check here?

manuelpolzhofer: I will add it

PR #275, opened 2020-01-26
chore: deployer refactor

lucasvo:
I've hardened our deployer a bit:
1. I'm removing all arguments from the deploy methods and moving them into the constructor of the deployer. This will make it easier to verify the parameters that it was deployed with
2. I'm making sure each deploy method can only be called once.
3. Because of these two changes, we can get rid of the deploy user. The deployer does not need to authenticate anyone anymore for calling a deploy method. 
4. I'm renaming RootAdmin to TinlakeRoot


src/lender/deployer.sol
```diff
@@ -0,0 +1,243 @@
...
+}
+
+contract SeniorTrancheFab {
+    string constant public symbol = "DROP";
+    string constant public name = "Tinlake Drop Token";
+    uint public ratePerSecond;
+
+    constructor (uint rate_) public {
+        ratePerSecond = rate_;
+    }
```
manuelpolzhofer: We want to re-use the Fabs on mainnet for different deployments. (with different rates)

I think each deployment will have a different token name

It should not be required to touch a solidity file if we want to deploy for another asset group.

src/root.sol
```diff
@@ -12,53 +12,54 @@
...
-        lenderDeployer.assessor().depend("pool", borrowerPricePool);
-    }
-
-    function relyModules() internal {
-        // distributor allowed to call
-        borrowerDeployer.shelf().rely(address(lenderDeployer.distributor()));
-    }
-
-    function completeDeployment() public auth {
+    function deploy() public auth {
```
manuelpolzhofer: I think we can remove the auth here.

src/borrower/test/mock/shelf.sol
```diff
@@ -4,7 +4,7 @@ import "../../../test/mock/mock.sol";
 
 contract ShelfMock is Mock {
 
-    // maybe not needed
+    // TODO: maybe not needed
```
manuelpolzhofer: can be removed. 
```suggestion

```

src/lender/deployer.sol
```diff
@@ -191,10 +191,10 @@ contract LenderDeployer is Auth {
         AuthLike(distributor).rely(root);
     }
 
-    function deploySeniorTranche() public  {
-        require(assessor != ZERO && senior == ZERO);
-        senior = seniorTrancheFab.newTranche(currency, assessor);
-        AuthLike(senior).rely(root);
+    function deployAssessor() public {
+        require(assessor == ZERO);
```
manuelpolzhofer: This should change with latest develop

src/borrower/deployer.sol
```diff
@@ -0,0 +1,215 @@
...
+
+    function deployCeiling() public {
+        require(ceiling == ZERO);
+        ceiling = ceilingFab.newCeiling();
+        AuthLike(ceiling).rely(root);
+    }
+
+    function deploy() public {
+        // ensures all required deploy methods were called
+        require(shelf != ZERO && collector != ZERO && pricePool != ZERO);
```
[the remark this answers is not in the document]
manuelpolzhofer: ahh, shelf transitive ensures it.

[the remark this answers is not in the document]
lucasvo: yes I removed the transitive checks.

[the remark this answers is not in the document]
lucasvo: (duplicate) checks

src/lender/deployer.sol
lucasvo: I agree, but I didn't want to blow up this even more. I think the added security for now is better than the flexibility we had. I will try to add it in a separate PR.

PR #274, opened 2020-01-26
fix: assessor calcMinJuniorAssetValue

manuelpolzhofer:
I think it was wrong because we refactored the ratio methods.

The method is called in the redeemApprove for junior
```javascript
        if (tranche == junior && safeSub(calcAssetValue(junior), currencyAmount) >= calcMinJuniorAssetValue()) {
            return true;	 
        }
```

Let's imagine we close the pool. Senior redeemed all the currency. 

For junior it would not be possible to redeem now

PR #276, opened 2020-01-26
test: added distributor system test

manuelpolzhofer:
please review #274 first


src/core/test/system/lender/assessor/default_interest.sol
```diff
@@ -17,7 +17,7 @@ pragma solidity >=0.5.12;
```
manuelpolzhofer: Dont look at this files. Nothing changed in interest tests.

I realized i switched the names in the test files.

`default_interest.sol` had the tests for Full invest interest
`full_invest_interest.sol` had the tests for default interest

PR #277, opened 2020-01-26
fix: no drip on first senior rate file && assessor requires

manuelpolzhofer:
- added require checks to the assessor

- we call `senior.file("rate", rate)` in the seniorFab.
- this caused a drip function  call which triggered an assessor call before the senior was set in the assessor
- the drip is not required


src/lender/tranche/senior_tranche.sol
```diff
@@ -52,12 +52,10 @@ contract SeniorTranche is Tranche, Interest {
 
     function file(bytes32 what, uint ratePerSecond_) public note auth {
          if (what ==  "rate") {
-            drip();
```
manuelpolzhofer: the most important change in this PR

ilinzweilin: I think we still need to drip in case file is used to update the rate

PR #280, opened 2020-01-26
add: tests for creditline contract

ilinzweilin:
- refactor: ceiling method: creditLine. Debt can grow bigger then credit Line when interest accrues. In this case credit line should return 0.
- added tests from credit line
- added creditline stup to deployer

### DIFF
--- a/src/borrower/deployer.sol
+++ b/src/borrower/deployer.sol
@@ -20,16 +20,27 @@ import { Shelf } from "./shelf.sol";
 import { Pile } from "./pile.sol";
 import { Collector } from "./collect/collector.sol";
 import { Principal } from "./ceiling/principal.sol";
+import { CreditLine } from "./ceiling/creditLine.sol";
 import { PushRegistry } from "tinlake-registry/registry.sol";
 import { PricePool } from "./price/pool.sol";
+
+contract DependLike {
+    function depend(bytes32, address) public;
+}
 
 contract AuthLike {
     function rely(address) public;
     function deny(address) public;
 }
 
-contract DependLike {
-    function depend(bytes32, address) public;
+contract TCeilingLike {
+    function ceiling(uint loan) public view returns(uint);
+    function values(uint) public view returns(uint);
+    function file(uint loan, uint currencyAmount) public;
+    function borrow(uint loan, uint currencyAmount) public;
+    function repay(uint loan, uint currencyAmount) public;
+    function rely(address) public;
+    function deny(address) public;
 }
 
 contract PileFab {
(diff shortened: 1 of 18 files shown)


src/borrower/deployer.sol
```diff
@@ -186,9 +193,14 @@ contract BorrowerDeployer {
         AuthLike(shelf).rely(root);
     }
 
-    function deployCeiling() public {
+    function deployCeiling(bytes32 ceiling_) public {
         require(ceiling == ZERO);
-        ceiling = ceilingFab.newCeiling();
+        if (ceiling_ == "default") {
+         ceiling = ceilingFab.newCeiling();
```
manuelpolzhofer: I would to it in the following way
```suggestion
         ceiling = ceilingFab.newCeiling(address(pile));
```

src/borrower/deployer.sol
```diff
@@ -186,9 +193,14 @@ contract BorrowerDeployer {
         AuthLike(shelf).rely(root);
     }
 
-    function deployCeiling() public {
+    function deployCeiling(bytes32 ceiling_) public {
```
manuelpolzhofer: Our deployer should have any optional parameter.

It would break the entire concept

src/borrower/deployer.sol
```diff
@@ -75,6 +76,13 @@ contract CeilingFab {
...
         return address(ceiling);
     }
+
+    function newCeiling(address pile) public returns (address) {
+        CreditLine ceiling = new CreditLine(pile);
+        ceiling.rely(msg.sender);
+        ceiling.deny(address(this));
+        return address(ceiling);
+    }
 }
```
manuelpolzhofer:
```javascript
contract CredlineCeilingFab {
  function newCeiling(address pile) public returns (address) {
        CreditLine ceiling = new CreditLine(pile);
        ceiling.rely(msg.sender);
        ceiling.deny(address(this));
        return address(ceiling);
    }

contract PrincipalCeilingFab {
  function newCeiling(address pile) public returns (address) {

        Principal ceiling = new Principal();
        ceiling.rely(msg.sender);
        ceiling.deny(address(this));
        return address(ceiling);
    }

```

manuelpolzhofer: lets do it in this way

manuelpolzhofer: `CeilingFab` should be abstract

src/borrower/deployer.sol
```diff
@@ -68,8 +74,17 @@ contract CollectorFab {
     }
 }
 
-contract CeilingFab {
-    function newCeiling() public returns (address) {
+contract CreditLineCeilingFab {
```
manuelpolzhofer: +1

PR #281, opened 2020-01-26
tests: added system tests for approveSupply and redeemSupply

manuelpolzhofer:


src/test/system/lender/ratio/two-tranches.sol
```diff
@@ -0,0 +1,150 @@
...
+        uint seniorInvestorAmount = 400 ether;
+        uint juniorInvestorAmount = 100 ether;
+
+        // minJuniorRatio 20%
+        uint minJuniorRatio = 2 * 10**26;
+        AssessorLike(assessor).file("minJuniorRatio" , minJuniorRatio);
+
+        supplyJunior(juniorInvestorAmount);
+        supplySenior(seniorInvestorAmount);
+
```
ilinzweilin: really good test.  I think it would make sense to pick a lower number for senior investment then 400 though (A value that is not max allowed senior investment). Because otherwise the additional supplyToSenior is already not permitted even before the default 🙈

manuelpolzhofer: I wanted to test the edge cases :)

PR #304, opened 2020-02-14
Chore - make deploy script more configurable

ilinzweilin:


bin/deploy.sh
```diff
@@ -30,21 +30,11 @@ source ./borrower/deploy.sh
 
 # finalize deployment
 message Finalize Deployment
-GOVERNANCE="0x$ETH_FROM"
 
 seth send $ROOT_CONTRACT 'prepare(address,address,address)' $LENDER_DEPLOYER $BORROWER_DEPLOYER $GOVERNANCE
```
manuelpolzhofer: if you remove it this line would not work anymore?

bin/deploy.sh
```diff
@@ -30,21 +30,11 @@ source ./borrower/deploy.sh
...
 seth send $ROOT_CONTRACT 'deploy()'
 
 success_msg "Tinlake Deployment Finished"
 success_msg "Deployment File: $(realpath $DEPLOYMENT_FILE)"
 
-#touch $DEPLOYMENT_FILE
-addValuesToFile $DEPLOYMENT_FILE <<EOF
```
manuelpolzhofer: I think we should add it the deployment json file to know for each deployment who is in control

bin/deploy.sh
ilinzweilin: it will, I have added it to the local config

bin/deploy.sh
ilinzweilin: it is added, but already in the setup_local_config_file

PR #309, opened 2020-02-18
fix - use safe math

ilinzweilin:
- adjust paths in deploy scripts
- use safeMath in operator and tests
- make allowance maps public
- remove dupl. depend call


bin/test/setup_local_config.sh
```diff
@@ -2,12 +2,12 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 BIN_DIR=${BIN_DIR:-$(cd "${0%/*}"&&pwd)}
 cd $BIN_DIR
 
-source ./../util/util.sh
+source $BIN_DIR/../util/util.sh
```
manuelpolzhofer: does it make a difference?

because I am doing `cd $BIN_DIR` above?

ilinzweilin: it does not complain that the files not exist when running for the first time

PR #310, opened 2020-02-19
chore: comments for pile, shelf, collector and tranche

manuelpolzhofer:
I started to add more comments.

please my make suggestions for spelling, grammar mistakes or additional sentences.


src/borrower/collect/collector.sol
```diff
@@ -72,26 +72,33 @@ contract Collector is DSNote, Auth {
...
+    /// therefore the ownership of the nft is transferred to the collector
     function seize(uint loan) external {
         uint debt = pile.debt(loan);
         require((threshold.get(loan) <= debt), "threshold-not-reached");
         shelf.claim(loan, address(this));
     }
 
+
+    /// a nft can be collected if the collector is the nft- owner
+    /// which implies the debt was above the loan threshold at certain point in time
```
charleenfei: which implies the debt was above the loan threshold at certain point in time and then seized by the collector

src/lender/tranche/tranche.sol
```diff
@@ -61,23 +62,28 @@ contract Tranche is DSNote, Auth {
...
+    /// supplies currency to the tranche and mints new tranche token in exchange
+    /// the exchange rate is defined by the ward contract
     function supply(address usr, uint currencyAmount, uint tokenAmount) external note auth {
         require(currency.transferFrom(usr, self, currencyAmount), "currency-transfer-failed");
         token.mint(usr, tokenAmount);
     }
 
+    /// redeems currency from a tranche in exchange for tranche tokens
+    /// burns the tranche token
+    /// the exchange rate is defined by the ward contract
```
charleenfei: ' exchange rate is defined by the ward contract' i don't think this is entirely accurate, it is determined by the tranche asset value and how much has been invested

src/lender/tranche/tranche.sol
```diff
@@ -61,23 +62,28 @@ contract Tranche is DSNote, Auth {
         return token.totalSupply();
     }
 
-    // -- Lender Side --
+    /// supplies currency to the tranche and mints new tranche token in exchange
+    /// the exchange rate is defined by the ward contract
```
charleenfei: see comment below

src/borrower/collect/collector.sol
manuelpolzhofer: 👍

src/lender/tranche/tranche.sol
manuelpolzhofer: yes, but from the perspective of tranche contract, it is defined by another contract.
how it is done, exactly might be confusing here for the reader

manuelpolzhofer: that was my thinking

src/borrower/collect/collector.sol
lucasvo: I think I would use 'An NFT' for this. The rules are not super clear cut on this but it rolls of the tongue more easily.

 https://www.quora.com/Why-is-an-used-instead-of-a-before-the-word-NGO-even-though-N-is-not-a-vowel?share=1

src/lender/tranche/tranche.sol
lucasvo: I agree with your point manuel. I would limit it to the functionality that is relevant to the code here.

src/borrower/collect/collector.sol
```diff
@@ -72,26 +72,34 @@ contract Collector is DSNote, Auth {
...
+    function depend(bytes32 contractName, address addr) external auth {
+        if (contractName == "distributor") distributor = DistributorLike(addr);
+        else if (contractName == "shelf") shelf = ShelfLike(addr);
+        else if (contractName == "pile") pile = PileLike(addr);
+        else if (contractName == "threshold") threshold = RegistryLike(addr);
         else revert();
     }
 
-    // --- Collector ---
+    /// sets the selling-price of an NFT
```
lucasvo: `selling-price` is not really a common noun. You could use `liquidation price`

src/borrower/collect/collector.sol
```diff
@@ -72,26 +72,34 @@ contract Collector is DSNote, Auth {
...
     function file(uint loan, address buyer, uint nftPrice) external auth {
         require(nftPrice > 0, "no-nft-price-defined");
         options[loan] = Option(buyer, nftPrice);
     }
 
+
+    /// if a loan debt is above the loan threshold the nft should be sold
```
lucasvo:
```suggestion
    /// if the loan debt is above the loan threshold the NFT should be seized, i.e. taken away from the borrower to be sold off at a later stage.
```

src/borrower/collect/collector.sol
lucasvo: Capitalize NFT and use "An NFT"

lucasvo: I would remove the "implies" part. You can say, to:
```suggestion
    /// The NFT needs to be `seized` first to transfer ownership to the collector.
```

src/borrower/pile.sol
```diff
@@ -24,32 +24,38 @@ import "tinlake-auth/auth.sol";
 // rate accumulators (chi values) for all interest rate categories. It calculates debt each
 // loan according to its interest rate category and pie value.
 contract Pile is DSNote, Auth, Interest {
-    // --- Data ---
+    /// stores all needed information of an interest rate group
```
lucasvo:
```suggestion
    // --- DATA ---
    
    /// stores all needed information of an interest rate group
```

src/borrower/pile.sol
```diff
@@ -24,32 +24,38 @@ import "tinlake-auth/auth.sol";
...
     struct Rate {
         uint   pie;                 // Total debt of all loans with this rate
         uint   chi;                 // Accumulated rates
         uint   ratePerSecond;       // Accumulation per second
         uint48 lastUpdated;         // Last time the rate was accumulated
     }
 
+    /// mapping for different interest rate groups
```
lucasvo:
```suggestion
    /// Interest Rate Groups are identified by a `uint` and stored in a mapping
```

src/borrower/pile.sol
```diff
@@ -24,32 +24,38 @@ import "tinlake-auth/auth.sol";
...
     constructor() public {
         wards[msg.sender] = 1;
+        /// pre-definition for loans without interest rates
         rates[0].chi = ONE;
         rates[0].ratePerSecond = ONE;
     }
 
-    // --- Public Debt Methods  ---
-    // increase the loan's debt by currencyAmount
+    /// increases the debt of a loan by a currencyAmount
```
lucasvo:
```suggestion
     // --- Public Debt Methods  --- 
    /// increases the debt of a loan by a currencyAmount
```

manuelpolzhofer:
I added some more comments

PR #315, opened 2020-03-10
fix: view for assessor methods

manuelpolzhofer:
- senior.debt() returns the latest debt but doesn't modify the state anymore

-  calcUpdatedTokenPrice, redeemApprove, supplyApprove update the state by calling drip before using the view methods

- if we use one of the assessor methods in a transaction we should always update debt before


src/lender/tranche/senior_tranche.sol
```diff
@@ -36,17 +36,21 @@ contract SeniorTranche is Tranche, Interest {
...
     }
 
+    function updatedDebt() external returns(uint) {
+        drip();
+        return safeAdd(borrowed, interest);
+    }
+
+    function debt() external view returns(uint) {
+        return safeAdd(borrowed, _calcInterest());
+    }
```
manuelpolzhofer: biggest change.

all other contracts should use `updatedDebt` only the assessor uses `debt` as view. (and might call `drip` before)

src/lender/assessor/base.sol
```diff
@@ -185,6 +199,7 @@ contract BaseAssessor is Math, Auth {
         if (tranche == junior || minJuniorRatio == 0) {
             return true;
         }
+        _drip();
```
manuelpolzhofer: supplyApprove should update the debt as well.

src/lender/assessor/base.sol
```diff
@@ -201,6 +216,8 @@ contract BaseAssessor is Math, Auth {
             return true;
         }
 
+        _drip();
```
manuelpolzhofer: same for redeem approve

src/lender/assessor/base.sol
```diff
@@ -81,22 +83,34 @@ contract BaseAssessor is Math, Auth {
...
+        return _calcSeniorAssetValue(poolValue, trancheReserve, _seniorDebt(), _juniorReserve());
     }
 
-    function calcTokenPrice(address tranche) public returns (uint) {
+    function calcTokenPrice(address tranche) public view returns (uint) {
         require(tranche  == junior || tranche == senior);
         return safeMul(_calcTokenPrice(tranche), tokenAmountForONE);
     }
+    
+    function calcUpdatedTokenPrice(address tranche) public returns (uint) {
```
manuelpolzhofer: all other contracts should use this method

src/lender/assessor/base.sol
ilinzweilin: I think it would make sense to make this function view as well

src/lender/assessor/base.sol
ilinzweilin: same here, having this function as a view would be really helpful

charleenfei: *note

src/lender/assessor/base.sol
charleenfei: *note

PR #319, opened 2020-03-15
feature - add descriptive messages

ilinzweilin:


bin/deploy.sh
```diff
@@ -19,6 +19,9 @@ success_msg "Correct Config File"
 # build contracts if needed
 build_contracts "./.."
 
+# create deployment folder
+mkdir $BIN_DIR/../deployments
```
manuelpolzhofer: let's check if the directory exists
```suggestion
[ ! -d "$BIN_DIR/../deployments" ] && mkdir $BIN_DIR/../deployments
```

lucasvo: Even simpler:
```suggestion
mkdir -p $BIN_DIR/../deployments
```

PR #327, opened 2020-03-26
chore: consistent file usage 

manuelpolzhofer:
implements suggestion 3 from the report.

**NOTE**
Breaking Changes for `tinlake.js`

### DIFF
--- a/src/test/system/users/admin.sol
+++ b/src/test/system/users/admin.sol
@@ -40,7 +40,7 @@ contract AdminUser {
     }
 
     function setCeiling(uint loan, uint principal) public {
-        ceiling.file(loan, principal);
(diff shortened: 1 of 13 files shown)


src/test/system/users/admin.sol
```diff
@@ -40,30 +40,30 @@ contract AdminUser {
     }
 
     function setCeiling(uint loan, uint principal) public {
-        ceiling.file(loan, principal);
+        ceiling.file("loan", loan, principal);
     }
 
     function doAdmit(address registry, uint nft, uint principal, address usr) public returns (uint) {
```
ilinzweilin: I don't think we need this test anymore. Looks like a v1 artifact.

PR #329, opened 2020-03-30
fix: path to env-check.sh

charleenfei:


bin/util/env-check.sh
```diff
@@ -1,6 +1,6 @@
 
 BIN_DIR=${BIN_DIR:-$(cd "${0%/*}"&&pwd)}
-source $BIN_DIR/util.sh
+source $BIN_DIR/util/util.sh
```
manuelpolzhofer: we are using it in the deploy script and for me, it's working.

Is it failing if you use it directly?

charleenfei: yes

ilinzweilin: for me it only works with Charly's fix

PR #334, opened 2020-03-31
feat: nftFeed for loan related values

manuelpolzhofer:
I opened a draft PR to better discuss the design of the NFTFeed.


src/borrower/nftfeed/nftfeed.sol
```diff
@@ -0,0 +1,126 @@
...
+        require(ceilingRatio[risk_] != 0, "ceiling for risk group not defined");
+        require(rate[risk_] != 0, "rate for risk group not defined");
+
+        risk[nftID_] = risk_;
+        nftValues[nftID_] = value;
+    }
+
+    // sets the loan rate in pile
+    // not possible for ongoing loans
+    function setPileRate(uint loan) external auth {
```
manuelpolzhofer: We can do it like this or refactor the code and the pile gets the rate from the NFTFeed. (more elegant but more work)

src/borrower/nftfeed/wrapper.sol
```diff
@@ -0,0 +1,34 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+pragma solidity >=0.5.3;
+
+import "./nftfeed.sol";
+
+contract NFTFeedWrapper is NFTFeed {
```
manuelpolzhofer: For not modifying the core contracts we can use a wrapper contract to implement the `ceiling` and ` threshold` interface

src/borrower/nftfeed/wrapper.sol
```diff
@@ -0,0 +1,34 @@
...
+
+    function borrow(uint loan, uint amount) external auth {
+        // ceiling check uses existing loan debt
+        require(ceiling(loan) >= safeAdd(pile.debt(loan), amount), "borrow-amount-too-high");
+    }
+
+    function repay(uint loan, uint amount) external auth {}
+
+    // Threshold Wrapper
+    // todo
```
manuelpolzhofer: Since the collector threshold is already a function call to an external contract. I won't implement a wrapper and modify the collector contract instead

src/borrower/nftfeed/nftfeed.sol
```diff
@@ -0,0 +1,126 @@
...
+        ceilingRatio[risk_] = ceilingRatio_;
+        rate[risk_]= rate_;
+    }
+
+    function file(bytes32 what, uint risk_, uint value_) public auth {
+        if (what == "threshold") { thresholdRatio[risk_]  = value_; }
+        else if (what == "ceiling") { ceilingRatio[risk_] = value_; }
+        else if (what == "rate") { rate[risk_]= value_;}
+        else revert("unknown parameter");
+    }
```
manuelpolzhofer: We could allow the admin to update individually ceiling, rate and threshold.

(If we allow it, we need three requires in the `update` method below  because not all values of a risk group are set)

src/borrower/nftfeed/nftfeed.sol
manuelpolzhofer: We need to be aware if we change the rate of a `risk group`  all ongoing loans of the risk group would still have the old rate in pile. Maybe that's okay, otherwise, we would have to call `changeRate` on all loans of the group.

If the pile would directly call the nftFeed for getting the rate. It would also be tricky if the rate changes (because it would require a recalculation of the `pie` (see change rate).

We could also say it is not possible to change the rate of a risk group instead move all members to a new risk group


We

manuelpolzhofer:
NFTFeed source code moved to own repository.

See: https://github.com/centrifuge/tinlake-nftfeed

PR #341, opened 2020-04-06
feat: fabDeployer contract 

manuelpolzhofer:
- contract code for fab deployer
- should we move the contract and the bash script into a own repo?


manuelpolzhofer:
moved PR into own deploy repo
https://github.com/centrifuge/tinlake-deploy/pull/7

PR #351, opened 2020-04-23
feat: prop operator

manuelpolzhofer:
- impl of prop operator
- tests for prop operator

**spec for prop operator** 
https://centrifuge.hackmd.io/iP40MNA2TlaYF5F_TeyzRQ
**modelling for prop operator** 
https://docs.google.com/spreadsheets/d/1u-3qjHJJthc87qGUMhxgr1ro75IMxQFhEU21RpJq_Vw/edit#gid=1024434668

### DIFF
--- a/src/lender/tranche/operator/proportional.sol
+++ b/src/lender/tranche/operator/proportional.sol
@@ -144,9 +144,8 @@ contract ProportionalOperator is Math, DSNote, Auth  {
             tokenAmount = maxTokenAmount;
         }
 
-    (uint currencyAmount, uint currencyRedeemed_,uint  principalRedeemed_) = calcRedeemCurrencyAmount(msg.sender, tokenAmount, maxTokenAmount);
-        currencyRedeemed[msg.sender] = currencyRedeemed_;
-        principalRedeemed[msg.sender] = principalRedeemed_;
+        uint currencyAmount = calcRedeemCurrencyAmount(msg.sender, tokenAmount, maxTokenAmount);
+
         require(assessor.redeemApprove(address(tranche), currencyAmount), "redeem-not-approved");
         tokenRedeemed[msg.sender] = safeAdd(tokenRedeemed[msg.sender], tokenAmount);
         tranche.redeem(msg.sender, currencyAmount, tokenAmount);
@@ -166,26 +165,23 @@ contract ProportionalOperator is Math, DSNote, Auth  {
     /// calculates the amount of currency a user can redeem for a specific token amount
     /// the used token price for the conversion can be different among users depending on their
     /// redeem history.
-    function calcRedeemCurrencyAmount(address usr, uint tokenAmount, uint maxTokenAmount) public view returns(uint, uint, uint) {
+    function calcRedeemCurrencyAmount(address usr, uint tokenAmount, uint maxTokenAmount) public returns(uint) {
         // solidity gas-optimized calculation avoiding local variable if possible
+        uint currencyAmount = rmul(tokenAmount, calcTokenPrice(usr));
 
         uint redeemRatio = rdiv(tokenAmount, maxTokenAmount);
 
-        // c is the delta between total currency returned and the portion the user has redeemed of it.
-        uint c = safeSub(totalCurrencyReturned, currencyRedeemed[usr]);
+        currencyRedeemed[usr] = safeAdd(rmul(safeSub(totalCurrencyReturned, currencyRedeemed[usr]),
+            redeemRatio), currencyRedeemed[usr]);
 
-        // p is the delta between total principal returned and the portion the user has redeemed of it.
-        uint p = safeSub(totalPrincipalReturned, principalRedeemed[usr]);
+        principalRedeemed[usr] = safeAdd(rmul(safeSub(totalPrincipalReturned, principalRedeemed[usr]),
+            redeemRatio), principalRedeemed[usr]);
 
-        return (
-        // calculates currencyAmount by multiplying the tokenAmount with the tokenPrice
-        rmul(tokenAmount, rdiv(c, p)),
+        return currencyAmount;
+    }
 
-        // updated currencyRedeemed of the user
-        safeAdd(rmul(c, redeemRatio), currencyRedeemed[usr]),
-
-        // updated principalRedeemed of the user
-        safeAdd(rmul(p, redeemRatio), principalRedeemed[usr])
-        );
+    function calcTokenPrice(address usr) public view returns (uint) {
+       return rdiv(safeSub(totalCurrencyReturned, currencyRedeemed[usr]),
+           safeSub(totalPrincipalReturned, principalRedeemed[usr]));
     }
 }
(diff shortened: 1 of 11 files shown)


src/lender/tranche/operator/proportional.sol
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    function file(bytes32 what, bool supplyAllowed_) public auth {
+        if(what == "supplyAllowed") {
+            supplyAllowed = supplyAllowed_;
+        }
+    }
+
+    function updateReturned(uint currencyReturned, uint principalReturned) public auth {
+        totalCurrencyReturned += currencyReturned;
```
manuelpolzhofer, line 86: maybe we should overwrite it here instead of adding. In case, it is set with the wrong values we can correct it

src/lender/tranche/operator/proportional.sol
```diff
@@ -0,0 +1,173 @@
...
+        totalPrincipalReturned += principalReturned;
+    }
+
+    /// defines the max amount of currency for supply
+    function approve(address usr, uint currencyAmount) external auth {
+        supplyMaximum[usr] = currencyAmount;
+        currentSupplyLimit[usr] = currencyAmount;
+    }
```
ilinzweilin, line 94: Problem: by overriding both values, we are loosing the entire investment history of the investor that is used in calcMaxRedeem. So if an investor has already supplied some DAI and his allowance is adjusted after, he won't be able to redeem his share as `supplyMaximum[usr] - currentSupplyLimit[usr]` will be 0. 
Possible Solutions: 
` currentSupplyLimit[usr] = currencyAmount - (supplyMaximum[usr] - currentSupplyLimit[usr])`
currencyLimit[usr] has to be an int which is not ideal, since it can turn into a negative value when `currencyAmount is 0 and investedAmount > 0`
so another possible solution would be having 2 different maps:
`currentSupplyLimit[usr]` which stores the currentSupplyLimit of the investor
`redeemedTokens[usr]` which stores the share in the pool that the user received --> can be used for calcMaxRedeem

src/lender/tranche/operator/proportional.sol
```diff
@@ -0,0 +1,173 @@
...
+
+        uint maxTokenAmount = calcMaxRedeemToken(usr);
+        require(tokenAmount <= maxTokenAmount, "tokenAmount higher than maximum");
+
+        uint redeemRatio = rdiv(tokenAmount, maxTokenAmount);
+
+        // c is the delta between total currency returned and the portion the user has redeemed of it.
+        uint c = safeSub(totalCurrencyReturned, currencyRedeemed[usr]);
+
+        // p is the delta between total currency returned and the portion the user has redeemed of it.
```
ilinzweilin, line 159: principal, not currency 🙈

src/lender/tranche/operator/proportional.sol
```diff
@@ -0,0 +1,173 @@
...
+        totalTrancheVolume = safeAdd(totalTrancheVolume, currencyAmount);
+
+        distributor.balance();
+    }
+
+    /// redeem is proportional allowed
+    function redeem(uint tokenAmount) external note {
+        distributor.balance();
+        (uint currencyAmount, uint currencyRedeemed_,uint  principalRedeemed_) = calcRedeemCurrencyAmount(msg.sender, tokenAmount);
+        currencyRedeemed[msg.sender] = currencyRedeemed_;
```
ilinzweilin, line 124: not better changing the state after calling tranche.redeem & receiving approval from assessor?

src/lender/tranche/operator/proportional.sol
manuelpolzhofer: simplified it in the lastest commit

src/lender/tranche/operator/proportional.sol
```diff
@@ -0,0 +1,171 @@
...
+        return safeSub(rmul(rdiv(totalPrincipalReturned, totalPrincipal), tokenReceived[usr]), tokenRedeemed[usr]);
+    }
+
+    /// calculates the amount of currency a user can redeem for a specific token amount
+    /// the used token price for the conversion can be different among users depending on their
+    /// redeem history.
+    function calcRedeemCurrencyAmount(address usr, uint tokenAmount) public view returns(uint, uint, uint) {
+        // solidity gas-optimized calculation avoiding local variable if possible
+
+        uint maxTokenAmount = calcMaxRedeemToken(usr);
```
ilinzweilin, line 147: //  maxTokenAmount that can be redeemed based on the investor's share in the pool

src/lender/tranche/operator/proportional.sol
```diff
@@ -0,0 +1,171 @@
...
+    DistributorLike public distributor;
+
+    // lender mappings
+    // each value in a own map for gas-optimization
+    mapping (address => uint) public supplyMaximum;
+    mapping (address => uint) public tokenReceived;
+    // helper we could also calculate it
+    mapping (address => uint) public tokenRedeemed;
+
+    // expressed in totalCurrencyReturned notation
```
ilinzweilin, line 51: // currency amount of which the investor's share in the pool has already been redeemed

src/lender/tranche/operator/proportional.sol
```diff
@@ -0,0 +1,171 @@
...
+    mapping (address => uint) public currencyRedeemed;
+
+    // expressed in totalPrincipalReturned notation
+    mapping (address => uint) public principalRedeemed;
```
ilinzweilin, line 55: suggestion: // principal amount of which the investor's share in the pool has already been redeemed

src/lender/tranche/operator/proportional.sol
ilinzweilin: maybe also calling the mappings currencyShareRedeemed & principalShareRedeemed

src/lender/tranche/operator/proportional.sol
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    /// calculates the amount of currency a user can redeem for a specific token amount
+    /// the used token price for the conversion can be different among users depending on their
+    /// redeem history.
+    function calcRedeemCurrencyAmount(address usr, uint tokenAmount) public view returns(uint, uint, uint) {
+        // solidity gas-optimized calculation avoiding local variable if possible
+
+        uint maxTokenAmount = calcMaxRedeemToken(usr);
+        require(tokenAmount <= maxTokenAmount, "tokenAmount higher than maximum");
```
lucasvo, line 147: @manuelpolzhofer I think we should make this method just use the maximum allowed in case tokenAmount > maxTokenAmount instead of a revert. Will make it simpler in the UI.

manuelpolzhofer: okay i will add

src/lender/tranche/operator/proportional.sol
```diff
@@ -0,0 +1,173 @@
...
+        if (supplyAllowed) {
+            return 0;
+        }
+
+        // todo figure out if it is cheaper to store the value every time a user redeems
+        uint previouslyRedeemed = rmul(rmul(rdiv(principalRedeemed[usr], totalPrincipalReturned),
+            supplyMaximum[usr]),rdiv(totalPrincipalReturned, totalTrancheVolume));
+
+        // considers the case if a user didn't supply the maximum amount possible
+        uint maxRedeemToken = rmul(rdiv(totalPrincipalReturned, totalTrancheVolume), safeSub(supplyMaximum[usr], currentSupplyLimit[usr]));
```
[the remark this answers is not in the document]
manuelpolzhofer: for simplification we only implemented `tokenPrice==ONE`

PR #355, opened 2020-05-01
fix: nft feed init order & prop operator init price

manuelpolzhofer:


src/lender/tranche/operator/proportional.sol
```diff
@@ -179,7 +179,11 @@ contract ProportionalOperator is Math, DSNote, Auth  {
         return currencyAmount;
     }
 
+    /// returns the tokenPrice denominated in RAD (10^27)
```
lucasvo, line 182: `ray`?

PR #374, opened 2020-08-12
feature - create reserve contract

ilinzweilin:


src/lender/reserve.sol
```diff
@@ -0,0 +1,60 @@
...
+        self = address(this);
+    }
+
+    function depend(bytes32 contractName, address addr) public auth {
+        if (contractName == "shelf") { shelf = ShelfLike(addr); }
+        else if (contractName == "currency") { currency = ERC20Like(addr); }
+        else revert();
+    }
+
+    function activatePool(bool active) public auth {
```
manuelpolzhofer, line 40: I think we don't need the activatePool method. i could just use `updateMaxCurrency(0)`. For the same effect

src/lender/reserve.sol
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    function balance() public{
+        require(poolActive, "pool-not-active");
+
+        (bool requestWant, uint currencyAmount) = shelf.balanceRequest();
+        if (requestWant) {
+            require(currencyAvailable >= currencyAmount, "not-enough-currency-reserve");
+            require(currency.transferFrom(self, address(shelf), currencyAmount), "currency-transfer-from-reserve-failed");
+            currencyAvailable = safeSub(currencyAvailable, currencyAmount);
```
manuelpolzhofer, line 55: Maybe a bit paranoid but lets to the safeSub before the send of the currency.

src/lender/test/reserve.t.sol
```diff
@@ -0,0 +1,134 @@
...
+        bool requestWant = false;
+        // fund shelf with currency amount smaller then repay amount
+        currency.mint(shelf_, safeSub(repayAmount, 1));
+        // borrow action: shelf requests currency
+        shelf.setReturn("balanceRequest", requestWant, repayAmount);
+        // shelf approve reserve to take currency
+        shelf.doApprove(currency_, reserve_, repayAmount);
+        reserve.balance();
+    }
+
```
manuelpolzhofer, line 132: Maybe we can add one more for reserve not enough funds and the edge case when it is equal

src/lender/reserve.sol
manuelpolzhofer: Because we only want to deactivate borrows. Repay are still okay

PR #388, opened 2020-08-17
feat: coordinator executeEpoch

manuelpolzhofer:
please review #373 first

In this PR:

- epochExecute method
- rebalancing of senior debt
- helper functions for epochExecute
- unit tests for epoch execute


src/lender/coordinator.sol
```diff
@@ -206,15 +214,49 @@ contract EpochCoordinator is Ticker, Auth  {
...
+
+    function calcFulfillment(uint amount, uint totalOrder) public view returns(uint percent) {
+        if(amount == 0 || totalOrder == 0) {
+            return 0;
+        }
+        return rdiv(amount, totalOrder);
+    }
+
     function executeEpoch(uint seniorRedeem, uint juniorRedeem, uint seniorSupply, uint juniorSupply) internal {
+        uint epochID = lastEpochExecuted+1;
```
ilinzweilin, line 247: shouldn't we check whether submission period is over: `require(now > minChallengePeriodEnd)`
also it just makes sense to execute if `lastEpochExecuted < closingEpoch`

manuelpolzhofer: the check is happening the the `public` executeEpoch method.

PR #389, opened 2020-08-18
feature: pile support fixed rates

ilinzweilin:
- add fixedRate var to rate groups
- adjust file function to set fixedRates on rate groups. If no fixedRate is set for a group value => 0
- in incDebt Add fixedBorrowFee: `currencyAmount = currencyAmount + currencyAmount*fixedRate` If fixedRate is zero, no fixedBorrowFee is added to currencyAmount
- Accrue interest on the fixedBorrowFee together with borrowAmount

### DIFF
--- a/src/borrower/pile.sol
+++ b/src/borrower/pile.sol
@@ -32,6 +32,7 @@ contract Pile is DSNote, Auth, Interest {
         uint   chi;                 // Accumulated rates
         uint   ratePerSecond;       // Accumulation per second
         uint48 lastUpdated;         // Last time the rate was accumulated
+        uint   fixedRate;           // fixed rate applied to each loan of the group
     }
 
     /// Interest Rate Groups are identified by a `uint` and stored in a mapping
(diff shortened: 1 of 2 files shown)


src/borrower/pile.sol
```diff
@@ -59,9 +60,10 @@ contract Pile is DSNote, Auth, Interest {
      // --- Public Debt Methods  ---
     /// increases the debt of a loan by a currencyAmount
     /// a change of the loan debt updates the rate debt and total debt
-    function incDebt(uint loan, uint currencyAmount) external auth note {
+    function incDebt(uint loan, uint currencyAmount) external auth note { 
         uint rate = loanRates[loan];
         require(now == rates[rate].lastUpdated, "rate-group-not-updated");
+        currencyAmount = safeAdd(currencyAmount, rmul(currencyAmount, rates[rate].fixedRate));
```
manuelpolzhofer, line 66: We could safe one addition if the fixed rate is in a different format:

For example 2% fixed fee instead of `0.02` we could express it as `1.02`
and therefore just multiply once

```suggestion
        currencyAmount =  rmul(currencyAmount, rates[rate].fixedRate));
```

src/borrower/pile.sol
```diff
@@ -133,18 +135,19 @@ contract Pile is DSNote, Auth, Interest {
     }
 
     // set/change the interest rate of a rate category
-    function file(bytes32 what, uint rate, uint ratePerSecond) external auth note {
+    function file(bytes32 what, uint rate, uint value) external auth note {
```
manuelpolzhofer, line 138:
```suggestion
    function file(bytes32 what, uint rateGroupID, uint rate) external auth note {
```

PR #390, opened 2020-08-18
feat: added improvement score and refactored coordinator

manuelpolzhofer:
**Features**
- split constraints into core constraints and pool constraints
- added in improvement scoring
- refactored coordinator


src/lender/coordinator.sol
```diff
@@ -125,97 +142,264 @@ contract EpochCoordinator is Ticker, Auth  {
...
+    function scoreDistanceReserve(Fixed27 memory newReserve_) public view returns (uint score) {
+        return rmul(1000, rdiv(ONE, abs(safeDiv(ONE, 2), newReserve_.value)));
+    }
+
+    function scoreImprovement(Fixed27 memory newSeniorRatio, Fixed27 memory currSeniorRatio,
+        uint newReserve_) public  returns(uint) {
+        (Fixed27 memory minSeniorRatio, Fixed27 memory maxSeniorRatio) = assessor.seniorRatioBounds();
+
+        // normalize reserve by defining maxReserve as ONE
+        Fixed27 memory normalizedNewReserve = Fixed27(rdiv(newReserve_, assessor.maxReserve()));
```
ilinzweilin, line 278: Maybe  this value can be computed in the normalizedNewReserve function, as it is not used in this function

src/lender/coordinator.sol
```diff
@@ -125,97 +142,264 @@ contract EpochCoordinator is Ticker, Auth  {
...
+        }
+
+        // gas optimized implementation
+        // abs of ratio can never be zero
+        uint score = rmul(10000, rdiv(ONE, abs(newSeniorRatio.value,
+            safeDiv(safeAdd(minSeniorRatio.value, maxSeniorRatio.value), 2))));
+
+        // ratio constraints and maxReserve are in the current state violated
+        // additional score
+        if (epochReserve >= assessor.maxReserve()) {
```
ilinzweilin, line 301: since fixing the ratio might actually put the reserve in a worse state then before, shouldn't we always compute the distanceScore of the reserve

src/lender/coordinator.sol
```diff
@@ -108,15 +116,24 @@ contract EpochCoordinator is Ticker, Auth  {
...
+            return;
+        }
+
         /// can orders be to 100% fulfilled
-        if (validate(orderSeniorRedeem, orderJuniorRedeem, orderSeniorSupply, orderJuniorSupply) == 0) {
+        if (validate(orderSeniorRedeem, orderJuniorRedeem,
+            orderSeniorSupply, orderJuniorSupply) == 0) {
 
-            executeEpoch(orderSeniorRedeem, orderJuniorRedeem, orderSeniorSupply, orderJuniorSupply);
+            executeEpoch(orderSeniorRedeem, orderJuniorRedeem,
```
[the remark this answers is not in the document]
manuelpolzhofer: The flow is as follow:
CloseEpoch -> SubmissionPeriod -> EpochExecuted

submissionPeriod is set to `true` in closeEpoch and is set to `false` in epochExecute.

We already check for `submissionPeriod == false` in the closeEpoch method.

It shouldn't be a problem.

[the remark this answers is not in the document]
manuelpolzhofer: wdyt?

src/lender/coordinator.sol
```diff
@@ -125,97 +142,264 @@ contract EpochCoordinator is Ticker, Auth  {
...
+        bestSubmission.juniorRedeem = juniorRedeem;
+        bestSubmission.juniorSupply = juniorSupply;
+        bestSubmission.seniorSupply = seniorSupply;
+
+        bestSubScore = score;
+    }
+
+    function submitSolution(uint seniorRedeem, uint juniorRedeem,
+        uint juniorSupply, uint seniorSupply) public returns(int) {
+
```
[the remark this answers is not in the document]
manuelpolzhofer: good idea

src/lender/coordinator.sol
manuelpolzhofer: If we calculate it here we only need to calculate it once.

If we calculate it in the method itself, the calculation would be performed twice

src/lender/coordinator.sol
manuelpolzhofer: Same for this issue. I will write it in the agenda for the meeting

ilinzweilin (approved):

ilinzweilin (approved):

PR #392, opened 2020-08-20
chore - integrate new nft feed interface

ilinzweilin:


src/borrower/deployer.sol
```diff
@@ -167,6 +168,9 @@ contract BorrowerDeployer {
 
             NFTFeedLike(nftFeed).init();
             DependLike(shelf).depend("subscriber", address(nftFeed));
+        } else {
+            // pool needs pile
+             DependLike(pricePool).depend("pile", address(pile));
```
manuelpolzhofer, line 173: In the if and in the else is the exact same command:

`DependLike(nftFeed).depend("pile", address(pile));` is the same as ` DependLike(pricePool).depend("pile", address(pile));`

src/borrower/fabs/navFeed.sol
```diff
@@ -0,0 +1,27 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+pragma solidity >=0.5.15 <0.6.0;
+
+import { NAVFeed } from "tinlake-nftfeed/feed.sol";
+
+contract NFTFeedFab {
```
manuelpolzhofer, line 20: I think we need a unique name.
```suggestion
contract NAVFeedFab {
```

src/borrower/fabs/navFeed.sol
```diff
@@ -0,0 +1,27 @@
...
+    function newNFTFeed() public returns (address) {
```
manuelpolzhofer, line 21: maybe we can call this one new feed
```suggestion
    function newFeed() public returns (address) {
```

manuelpolzhofer (approved):

PR #408, opened 2020-08-27
chore: better improvement function in coordinator

manuelpolzhofer:
please review #404 first

**New improvement scoring**
- own variables for bestRatioScore and bestReserveScore
- first compare ratioScore with bestRatioScore only if equal check reserve score
- if ratio is in healthy region give always max points


src/lender/coordinator.sol
```diff
@@ -243,65 +248,79 @@ contract EpochCoordinator is Ticker, Auth, DataTypes  {
         Fixed27 memory currSeniorRatio = Fixed27(calcSeniorRatio(epochSeniorAsset,
             epochNAV, epochReserve));
 
-        if(bestSubScore == 0) {
+        int err = 0;
+        uint impScoreRatio = 0;
+        uint impScoreReserve = 0;
+
+        if(bestRatioImprovement == 0) {
```
ilinzweilin, line 255:
```suggestion
        if (bestRatioImprovement == 0) {
```

ilinzweilin (approved):
👍

PR #410, opened 2020-08-27
chore: skip epoch if no orders

manuelpolzhofer:
- we can skip an epoch if there a no orders
- previously we calculated an updated NAV which was not required

@ilinzweilin did i miss something if an epoch is never updated/skipped in the tranche?


manuelpolzhofer:
We need to change this PR to the new tranche interfaces with epochClose and epochUpdate

manuelpolzhofer:
I will close this PR and add the logic to #415

PR #415, opened 2020-09-01
feat: added adapter interfaces

manuelpolzhofer:
please review #413 first

I added the methods which are required for the adapter

**Reserve**
- `function deposit(uint amount) public auth`
- `function payout(uint amount) public auth`

**Assessor**
- `increaseSeniorAsset(uint amount, uint newSeniorRatio) public auth`
- `decreaseSeniorAsset(uint amount, uint newSeniorRatio) public auth`

**Tranche**
- `mint(address usr, uint amount) public auth`

### DIFF
--- a/src/lender/assessor.sol
+++ b/src/lender/assessor.sol
@@ -32,25 +32,22 @@ interface TrancheLike {
 
 contract Assessor is Auth, DataTypes, Interest  {
     // senior ratio from the last epoch executed
-    Fixed27 public seniorRatio;
-
-    uint public seniorDebt_;
-    uint public seniorBalance_;
-
-    // system parameter
+    Fixed27        public seniorRatio;
+    uint           public seniorDebt_;
+    uint           public seniorBalance_;
 
     // interest rate per second for senior tranche
-    Fixed27 public seniorInterestRate;
-    uint public lastUpdateSeniorInterest;
+    Fixed27         public seniorInterestRate;
+    uint            public lastUpdateSeniorInterest;
 
-    Fixed27 public maxSeniorRatio;
-    Fixed27 public minSeniorRatio;
+    Fixed27         public maxSeniorRatio;
+    Fixed27         public minSeniorRatio;
 
-    uint public maxReserve;
+    uint            public maxReserve;
 
-    TrancheLike seniorTranche;
-    TrancheLike juniorTranche;
-    NAVFeedLike navFeed;
+    TrancheLike     public seniorTranche;
+    TrancheLike     public juniorTranche;
+    NAVFeedLike     public navFeed;
 
     constructor() public {
         wards[msg.sender] = 1;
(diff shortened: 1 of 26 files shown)


src/lender/tranche.sol
```diff
@@ -245,11 +244,11 @@ contract Tranche is Math, Auth, FixedPoint {
         adjustCurrencyBalance(epochID, epochSupplyCurrency, epochRedeemCurrency);
 
         totalSupply = safeAdd(safeSub(totalSupply, epochSupplyCurrency), rmul(epochSupplyCurrency, safeSub(ONE, epochs[epochID].supplyFulfillment.value)));
-        totalRedeem = safeAdd(safeSub(totalRedeem, epochRedeemCurrency), rmul(epochRedeemCurrency, safeSub(ONE, epochs[epochID].redeemFulfillment.value)));
+        totalRedeem = safeAdd(safeSub(totalRedeem, redeemInToken), rmul(redeemInToken, safeSub(ONE, epochs[epochID].redeemFulfillment.value)));
```
manuelpolzhofer, line 247: This was previously wrong

src/lender/coordinator.sol
```diff
@@ -120,6 +121,16 @@ contract EpochCoordinator is Ticker, Auth, FixedPoint  {
...
         (uint orderSeniorSupply, uint orderSeniorRedeem) = seniorTranche.closeEpoch();
 
+        //  if no orders exist epoch can be executed without validation
+        if (orderSeniorRedeem == 0 && orderJuniorRedeem == 0 &&
+        orderSeniorSupply == 0 && orderJuniorSupply == 0) {
+
+            juniorTranche.epochUpdate(0, 0, 0, orderJuniorSupply, orderJuniorRedeem);
+            seniorTranche.epochUpdate(0, 0, 0, orderSeniorSupply, orderSeniorRedeem);
+            lastEpochExecuted = safeAdd(lastEpochExecuted, 1);
+            return;
```
ilinzweilin, line 131: Even though there are no orders the capital might still need rebalancing (which only happens when execute epoch is called), due to the increase of the senior debt & the new NAV

src/lender/assessor.sol
```diff
@@ -81,17 +82,17 @@ contract Assessor is Auth, FixedPoint, Interest  {
...
         // we use the approximated NAV here because during the submission period
         // new loans might have been repaid in the meanwhile which are not considered in the epochNAV
         seniorDebt_ = rmul(navFeed.approximatedNAV(), seniorRatio_);
-        seniorBalance_ = safeSub(seniorAsset, seniorDebt_);
+        seniorBalance_ = safeSub(seniorAsset_, seniorDebt_);
+    }
 
+    function updateSeniorAsset(uint seniorRatio_) external auth {
+        dripSeniorDebt();
+        _rebalance(safeAdd(seniorDebt_, seniorBalance_), seniorRatio_);
```
ilinzweilin, line 95: Would be helpful to rename the state vars seniorDebt_ & seniorBalance_ -> can otherwise be confused with function variables.

ilinzweilin (approved):
coolio

src/lender/assessor.sol
manuelpolzhofer: we have a function called `seniorDebt()` thats the reason why i have this convention

src/lender/coordinator.sol
manuelpolzhofer: thats also true. We are trying to avoid to recalculating the NAV if its not necessary. 

We will refactor this part anyway with minimum epoch concept

PR #422, opened 2020-09-03
feature - add restricted token contract

ilinzweilin:
As dicussed with @lucasvo 
- the token restriction only applies to receiving the tokens
- we might extend the restriction to transferring the tokens in the future

Todo -> next pr
- if no memberlist provided to the token contract, behave like standard erc20


src/lender/token/restricted.sol
```diff
@@ -0,0 +1,47 @@
...
+    modifier checkMember(address usr) { memberlist.member(usr); _; }
+    
+    function hasMember(address usr) public returns (bool) {
+        // will fail if user not member
+        memberlist.member(usr);
+    }
+
+    constructor(string memory symbol_, string memory name_) ERC20(symbol, name) public {
+        wards[msg.sender] = 1;
+    }
```
manuelpolzhofer, line 36: I think the constructor is not required. The existing one adds the ward as well

src/lender/token/memberlist.sol
```diff
@@ -0,0 +1,39 @@
...
+import "tinlake-auth/auth.sol";
+
+contract Memberlist is Math, Auth {
+
+    uint constant minimumDelay = 7 days; 
+
+    // -- Members--
+    mapping (address => uint) public members;
+    function updateMember(address usr, uint validUntil) public auth { 
+        require((safeAdd(now, minimumDelay)) < validUntil);
```
manuelpolzhofer, line 28: We should also add a comment here explaining it
```suggestion
        require((safeAdd(block.timestamp, minimumDelay)) < validUntil);
```

```

File: src/lender/token/restricted.sol [hunk 3]
```diff
@@ -0,0 +1,47 @@
...
+
+    function depend(bytes32 contractName, address addr) public auth {
+        if (contractName == "memberlist") { memberlist = MemberlistLike(addr); }
+        else revert();
+    }
+
+    function transferFrom(address from, address to, uint wad) checkMember(to) public returns (bool) {
```
manuelpolzhofer, line 43: We need to add a check for `from` as well.

Let's say I am in the list receive tokens. Later I get removed.  I could still send the tokens to another address afterwards.

manuelpolzhofer: From the spec: `The token must only allow transfers to and from members in the list.`

src/lender/token/restricted.sol
```diff
@@ -0,0 +1,47 @@
...
+
+pragma solidity >=0.5.15 <0.6.0;
+
+import "tinlake-erc20/erc20.sol";
+
+contract MemberlistLike {
+    function member(address) public;
+}
+
+contract RestrictedToken is ERC20 {
```
manuelpolzhofer, line 24: Should we have a flag to activate and deactivate the memberlist?

src/lender/token/memberlist.sol
```diff
@@ -0,0 +1,39 @@
...
+        members[usr] = validUntil;
+     }
+
+    constructor() public {
+        wards[msg.sender] = 1;
+    }
+   
+    function member(address usr) public {
```
manuelpolzhofer, line 36:
```suggestion
    function member(address usr) public view {
```
Why not `isMember`?

src/lender/token/memberlist.sol
```diff
@@ -0,0 +1,39 @@
...
+        require((members[usr] >= now), "not-allowed-to-hold-token");
```
manuelpolzhofer, line 37: Why not returning `true` or `false` here?

It is hard to build logic on top if it would fail here. 
```
if member {
  do x 
} else
{
do y
}
```

src/lender/token/restricted.sol
```diff
@@ -0,0 +1,47 @@
...
+
+    MemberlistLike public memberlist; 
+    modifier checkMember(address usr) { memberlist.member(usr); _; }
+    
+    function hasMember(address usr) public returns (bool) {
```
manuelpolzhofer, line 29:
```suggestion
    function hasMember(address usr) public view returns (bool) {
```

src/lender/token/restricted.sol
```diff
@@ -0,0 +1,47 @@
...
+        // will fail if user not member
+        memberlist.member(usr);
```
manuelpolzhofer, line 31: We should return and not fail in my opinion

src/test/system/base_system.sol
```diff
@@ -143,8 +136,11 @@ contract BaseSystemTest is TestSetup, Math, DSTest {
 
     function createLoanAndWithdraw(address usr, uint nftPrice, uint riskGroup) public returns (uint, uint) {
         (uint loanId, uint tokenId) = createLoanAndBorrow(usr, nftPrice, riskGroup);
+        emit log_named_uint("gas",12);
```
manuelpolzhofer, line 139:
```suggestion

```

src/lender/token/restricted.sol
ilinzweilin: yes, that's the plan. Just did not make it part of this pr

manuelpolzhofer (approved):
looks good.

we can discuss the change of the member interface with Lucas and potentially change it in another PR

ilinzweilin:
opened new pr #429

PR #437, opened 2020-09-08
fix: refactor tranche fab for gas limit in block

charleenfei:


src/lender/deployer.sol
```diff
@@ -105,20 +107,23 @@ contract LenderDeployer is FixedPoint {
         juniorSymbol =juniorSymbol_;
     }
 
-    function init(uint minSeniorRatio_, uint maxSeniorRatio_, uint maxReserve_, uint challengeTime_, uint seniorInterestRate_) public {
+    function init(uint minSeniorRatio_, uint maxSeniorRatio_, uint maxReserve_, uint challengeTime_, uint seniorInterestRate_, MemberlistFab memberlistFab_) public {
```
manuelpolzhofer, line 110: Can we move the token name and symbol params to the init method?

Then we would have a pattern. fabs in the constructor and all other config params in init

src/lender/deployer.sol
```diff
@@ -85,40 +87,41 @@ contract LenderDeployer is FixedPoint {
...
         minSeniorRatio = Fixed27(minSeniorRatio_);
         maxSeniorRatio = Fixed27(maxSeniorRatio_);
         maxReserve = maxReserve_;
         seniorInterestRate = Fixed27(seniorInterestRate_);
 
+        // token names
+        seniorName = seniorName_;
+        seniorSymbol = seniorSymbol_;
+        juniorName = juniorName_;
+        juniorSymbol =juniorSymbol_;
```
manuelpolzhofer, line 116:
```suggestion
        juniorSymbol = juniorSymbol_;
```

manuelpolzhofer (approved):
LGTM

PR #445, opened 2020-09-09
fix: end epoch bug in tranche

manuelpolzhofer:
The problem occurs if not all epochs are disbursed in one run by using the endEpoch parameter.

The disburse starts with `orderedInEpoch` and the method would disburse all including endEpoch.

Afterwards orderedInEpoch is set to endEpoch. Which was wrong because the next disburse would start with previous last disburse epoch.

**Correct Behaviour**
`orderedInEpoch` should be updated with `endEpoch + 1`.

### DIFF
--- a/src/lender/tranche.sol
+++ b/src/lender/tranche.sol
@@ -200,18 +200,23 @@ contract Tranche is Math, Auth, FixedPoint {
     // the disburse function can be used after an epoch is over to receive currency and tokens
     function disburse(address usr,  uint endEpoch) public auth returns (uint payoutCurrencyAmount, uint payoutTokenAmount, uint remainingSupplyCurrency, uint remainingRedeemToken) {
         require(users[usr].orderedInEpoch <= epochTicker.lastEpochExecuted());
+
+        uint lastEpochExecuted = epochTicker.lastEpochExecuted();
+
+        if (endEpoch > lastEpochExecuted) {
+            // it is only possible to disburse epochs which are already over
+            endEpoch = lastEpochExecuted;
+        }
+
         (payoutCurrencyAmount, payoutTokenAmount,
          remainingSupplyCurrency, remainingRedeemToken) = calcDisburse(usr, endEpoch);
         users[usr].supplyCurrencyAmount = remainingSupplyCurrency;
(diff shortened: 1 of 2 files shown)


src/lender/tranche.sol
```diff
@@ -200,18 +200,23 @@ contract Tranche is Math, Auth, FixedPoint {
...
-        // which allows to change the order and therefore receive it back
+        // if lastEpochExecuted is disbursed orderInEpoch is at current epoch again
+        // which allows to change the order and therefore the possibility to remove it
         // this is only possible if all previous epochs are disbursed (no orders reserved)
-        if (endEpoch == epochTicker.lastEpochExecuted()) {
-            users[usr].orderedInEpoch = epochTicker.currentEpoch();
-        } else {
-            users[usr].orderedInEpoch = endEpoch;
-        }
+        users[usr].orderedInEpoch = safeAdd(endEpoch, 1);
```
manuelpolzhofer, line 218: The `if condition` is not required anymore because if endEpoch == lastEpochExecuted an increase would result in currentEpoch.

src/lender/tranche.sol
```diff
@@ -200,18 +200,23 @@ contract Tranche is Math, Auth, FixedPoint {
     // the disburse function can be used after an epoch is over to receive currency and tokens
     function disburse(address usr,  uint endEpoch) public auth returns (uint payoutCurrencyAmount, uint payoutTokenAmount, uint remainingSupplyCurrency, uint remainingRedeemToken) {
         require(users[usr].orderedInEpoch <= epochTicker.lastEpochExecuted());
+
+        uint lastEpochExecuted = epochTicker.lastEpochExecuted();
+
+        if (endEpoch > lastEpochExecuted) {
+            // it is only possible to disburse epochs which are already over
+            endEpoch = lastEpochExecuted;
+        }
```
manuelpolzhofer, line 209: This was another problem because this check previously only happened in the calcDisburse method.

It would allow an investor to use a endEpoch higher than current epoch the disburse calculation would fine but we would overwrite this variable in `line 213`

ilinzweilin (approved):

PR #446, opened 2020-09-10
chore: order allowed checks for supplyOrder and redeemOrder

manuelpolzhofer:
- more comments in NAV feed and tranche

**Current Impl**
A user can place an order if one of the following is true
- `users[usr].orderedInEpoch == 0`
- `users[usr].orderedInEpoch  == currentEpoch`

**New Impl**
A user can place an order if one of the following is true
-  `(users[usr].supplyCurrencyAmount == 0 && users[usr].redeemTokenAmount == 0)`
-  `users[usr].orderedInEpoch  == currentEpoch`

Currently we never set orderedInEpoch back to 0. If a user supply/redeem and disburse and want to re-supply after a couple of epochs. We currently would require to call disburse once to set orderedInEpoch back to currentEpoch. Before an order can be placed. Which is counter intuitive because an order was not placed.

We could add a check in disburse if remaining supply and redeem are empty to set orderedInEpoch back to 0 but I think its more elegant to change the require and not rely on orderedInEpoch == 0 anymore.

### DIFF
--- a/src/lender/tranche.sol
+++ b/src/lender/tranche.sol
@@ -74,6 +74,12 @@ contract Tranche is Math, Auth, FixedPoint {
 
     bool public waitingForUpdate = false;
 
+    modifier orderAllowed(address usr) {
+        require((users[usr].supplyCurrencyAmount == 0 && users[usr].redeemTokenAmount == 0)
+        || users[usr].orderedInEpoch == epochTicker.currentEpoch(), "disburse required");
+        _;
+    }
+
     constructor(address currency_, address token_) public {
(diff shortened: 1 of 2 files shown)


src/lender/tranche.sol
```diff
@@ -74,6 +74,12 @@ contract Tranche is Math, Auth, FixedPoint {
 
     bool public waitingForUpdate = false;
 
+    modifier orderAllowed(address usr) {
```
ilinzweilin, line 77: 👍

ilinzweilin (approved):

PR #451, opened 2020-09-16
chore: cleanup and tinlake math lib upgrade

manuelpolzhofer:
- removed todos from codebase
- removed unused legacy code from v0.2
- added additional unit test for tranche 
- upgraded math lib for having revert error messages in math operations


src/borrower/test/mock/pile.sol
```diff
@@ -8,7 +8,7 @@ contract PileMock is Mock {
     }
 
     /// returns the total debt of a interest rate group
-    function rateDebt(uint rate) public view returns (uint) {
+    function rateDebt(uint ) public view returns (uint) {
```
ilinzweilin, line 11:
```suggestion
    function rateDebt(uint) public view returns (uint) {
```

ilinzweilin (approved):

PR #453, opened 2020-09-16
chore - add more nav feed tests

ilinzweilin:
- only allow maturity date changes if no outstandign loan exists
- set collateral fv to 0, in case loan fully repaid
- consider fixed fees when calculating FV on borrow events

Todo next pr: system test including loan having fixed fees


src/borrower/feed/navfeed.sol
```diff
@@ -249,6 +256,12 @@ contract NAVFeed is BaseNFTFeed, Interest, Buckets, FixedPoint {
         bytes32 nftID_ = nftID(loan);
         uint maturityDate_ = maturityDate[nftID_];
 
+
+        // no fv decrease calculation needed if maturaity date is in the past
```
manuelpolzhofer, line 260: Maybe we can add a comment here that buckets in the past are not considered

src/borrower/feed/navfeed.sol
```diff
@@ -262,9 +275,10 @@ contract NAVFeed is BaseNFTFeed, Interest, Buckets, FixedPoint {
         if (debt != 0) {
             fv = calcFutureValue(loan, debt, maturityDate_, recoveryRatePD[risk[nftID_]].value);
             buckets[maturityDate_].value = safeAdd(buckets[maturityDate_].value, fv);
-            futureValue[nftID_] = fv;
         }
 
+        futureValue[nftID_] = fv;
```
manuelpolzhofer, line 280: We need to update it to potentially allow partial borrow, correct?

manuelpolzhofer (approved):
LGTM

src/borrower/feed/navfeed.sol
ilinzweilin: yes, exactly

PR #457, opened 2020-09-25
fix: safeTransfer for tranche & calcDisburse rounding off

manuelpolzhofer:
- if the tranche doesn't have enough currency or token instead of failing  the user should receive the maximum possible amount
- calcDisburse should always round off instead of round up instead of depending if the last value is `> 5` or `< 5`

### DIFF
--- a/src/lender/test/tranche.t.sol
+++ b/src/lender/test/tranche.t.sol
@@ -429,7 +429,7 @@ contract TrancheTest is DSTest, Math, FixedPoint {
         // the disburse method should always round off
         ( , uint payoutTokenAmount,, ) =  tranche.calcDisburse(self);
 
-        // rdiv would round up but calc disburse should round off
+        // rdiv would round up in the 20/3 case but calc disburse should always round off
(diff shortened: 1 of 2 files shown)


src/lender/test/tranche.t.sol
```diff
@@ -415,4 +415,22 @@ contract TrancheTest is DSTest, Math, FixedPoint {
...
+        uint supplyFulfillment_ = ONE;
+        uint redeemFulfillment_ = ONE;
+        uint tokenPrice_ = 3 * 10 ** 27;
+
+        closeAndUpdate(supplyFulfillment_, redeemFulfillment_, tokenPrice_);
+
+        // the disburse method should always round off
+        ( , uint payoutTokenAmount,, ) =  tranche.calcDisburse(self);
+
+        // rdiv would round up but calc disburse should round off
```
manuelpolzhofer, line 432:
```suggestion
        // rdiv would round up in the 20/3 case but calc disburse should always round off
```

lucasvo (approved):

ilinzweilin (approved):

PR #461, opened 2020-09-30
adding permit shortcut

lucasvo:


manuelpolzhofer (commented):
integration tests would be nice

lucasvo:
@manuelpolzhofer it's extremely complicated with hevm unfortunately... It doesn't allow us to mock ecrecover. I can create a mock permit though not sure how much that will help though.

PR #462, opened 2020-09-30
feat: recovery transfer for tranche

manuelpolzhofer:
- recovery transfer allows governance to intervene if tokens get stuck in the tranche contract
- long term method should be removed but for the first deployments it is an additional security mechanism to avoid locked funds

### DIFF
--- a/src/lender/test/tranche.t.sol
+++ b/src/lender/test/tranche.t.sol
@@ -28,7 +28,7 @@ contract Hevm {
 
 contract User {
     function authTransfer(Tranche tranche, address erc20, address usr, uint amount) public {
-        tranche.recoveryTransfer(erc20, usr, amount);
+        tranche.authTransfer(erc20, usr, amount);
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/lender/tranche.sol
```diff
@@ -333,4 +333,9 @@ contract Tranche is Math, Auth, FixedPoint {
             safePayout(diff);
         }
     }
+
+    // recovery transfer can be used by governance to recover funds if tokens are stuck
+    function recoveryTransfer(address erc20, address usr, uint amount) public auth {
```
lucasvo, line 338:
```suggestion
    function authTransfer(address erc20, address usr, uint amount) public auth {
```

src/lender/test/tranche.t.sol
```diff
@@ -21,12 +21,17 @@ import "tinlake-math/math.sol";
...
 import "../../test/simple/token.sol";
 import "../test/mock/reserve.sol";
-import "../../../lib/tinlake-erc20/src/erc20.sol";
 
 contract Hevm {
     function warp(uint256) public;
 }
 
+contract User {
+    function recoveryTransfer(Tranche tranche, address erc20, address usr, uint amount) public {
```
lucasvo, line 30:
```suggestion
    function authTransfer(Tranche tranche, address erc20, address usr, uint amount) public {
```

lucasvo (approved):
Just naming.

PR #468, opened 2020-10-01
chore - add fixed fee system test

ilinzweilin:
- fix for totalRedeem & totalSupply rounding error
- added more tests for borrowing with fixed fees
- bumbed tinlake-math submodule version. I think it got downgraded in one of the prev commits


src/lender/tranche.sol
```diff
@@ -339,10 +339,12 @@ contract Tranche is Math, Auth, FixedPoint {
         ERC20Like(erc20).transferFrom(self, usr, amount);
     }
 
+    // due to rounding in token & currency conversions currency & token balances might be off by 1 wei with the totalSupply/totalRedeem amounts.
+    // in order to prevent reverts 0 is returned when amountto be substracted is bigger then the total value.
```
lucasvo, line 343:
```suggestion
    // in order to prevent an underflow error, 0 is returned when amount to be subtracted is bigger then the total value.
```

lucasvo (approved):

src/lender/tranche.sol
```diff
@@ -111,7 +111,7 @@ contract Tranche is Math, Auth, FixedPoint {
 
         users[usr].supplyCurrencyAmount = newSupplyAmount;
 
-        totalSupply = safeAdd(safeSub(totalSupply, currentSupplyAmount), newSupplyAmount);
+        totalSupply = safeAdd(safeTotalSub(totalSupply, currentSupplyAmount), newSupplyAmount);
```
manuelpolzhofer, line 114: I think here never a problem can occur. `safeTotalSub` is not required here.

src/lender/tranche.sol
```diff
@@ -258,8 +258,8 @@ contract Tranche is Math, Auth, FixedPoint {
         adjustCurrencyBalance(epochID, epochSupplyOrderCurrency, epochRedeemOrderCurrency);
 
         // the unfulfilled orders (1-fulfillment) is automatically ordered
-        totalSupply = safeAdd(safeSub(totalSupply, epochSupplyOrderCurrency), rmul(epochSupplyOrderCurrency, safeSub(ONE, epochs[epochID].supplyFulfillment.value)));
-        totalRedeem = safeAdd(safeSub(totalRedeem, redeemInToken), rmul(redeemInToken, safeSub(ONE, epochs[epochID].redeemFulfillment.value)));
+        totalSupply = safeAdd(safeTotalSub(totalSupply, epochSupplyOrderCurrency), rmul(epochSupplyOrderCurrency, safeSub(ONE, epochs[epochID].supplyFulfillment.value)));
```
manuelpolzhofer, line 261: The problem can't occur here. `safeTotalSub` is not required.

src/lender/tranche.sol
```diff
@@ -258,8 +258,8 @@ contract Tranche is Math, Auth, FixedPoint {
...
+        totalRedeem = safeAdd(safeTotalSub(totalRedeem, redeemInToken), rmul(redeemInToken, safeSub(ONE, epochs[epochID].redeemFulfillment.value)));
```
manuelpolzhofer, line 262: Here is the only place where the rounding problem can happen because of `redeemInToken`.

src/lender/tranche.sol
```diff
@@ -338,4 +338,13 @@ contract Tranche is Math, Auth, FixedPoint {
...
     }
+
+    // due to rounding in token & currency conversions currency & token balances might be off by 1 wei with the totalSupply/totalRedeem amounts.
+    // in order to prevent an underflow error, 0 is returned when amount to be subtracted is bigger then the total value.
+    function safeTotalSub(uint total, uint amount) internal returns (uint) {
+        if (total < amount) {
+            return 0;
+        }
+        return safeSub(total, amount);
+    }
```
manuelpolzhofer, line 349: I would remove this method and fix the following:

line: 252 don't use `rdiv`
```javascript
        if(tokenPrice_ > 0) {
            supplyInToken = rdiv(epochSupplyOrderCurrency, tokenPrice_);
            redeemInToken = safeDiv(safeMul(epochRedeemOrderCurrency,ONE), tokenPrice_);
        }
```

manuelpolzhofer (commented):
Not using rdiv in line 252 would solve the root of the problem and is more gas efficient

src/lender/tranche.sol
ilinzweilin, line 114: This is exactly where the problem occured on kovan :D

manuelpolzhofer: I will take a deeper look

manuelpolzhofer (approved):
I think it makes sense to use safeTotalSub to prevent potential unknown rounding errors even if we believe they can't occur.

After more fuzzing testing we potential can remove it.

PR #490, opened 2020-11-26
fix: no rebalancing if epoch has no orders

manuelpolzhofer:
- if a epoch was closed without any orders we didn't call the assessor for the rebalancing
- if the NAV would change in the meanwhile this would cause incorrect senior interest calculation

### DIFF
--- a/src/lender/coordinator.sol
+++ b/src/lender/coordinator.sol
@@ -176,19 +176,24 @@ contract EpochCoordinator is Auth, Math, FixedPoint  {
         (uint orderJuniorSupply, uint orderJuniorRedeem) = juniorTranche.closeEpoch();
         (uint orderSeniorSupply, uint orderSeniorRedeem) = seniorTranche.closeEpoch();
 
-        //  if no orders exist epoch can be executed without validation
-        if (orderSeniorRedeem == 0 && orderJuniorRedeem == 0 &&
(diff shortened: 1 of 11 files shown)


src/lender/coordinator.sol
```diff
@@ -176,20 +176,25 @@ contract EpochCoordinator is Auth, Math, FixedPoint  {
...
         //  if no orders exist epoch can be executed without validation
         if (orderSeniorRedeem == 0 && orderJuniorRedeem == 0 &&
         orderSeniorSupply == 0 && orderJuniorSupply == 0) {
 
-            juniorTranche.epochUpdate(currentEpoch, 0, 0, 0, orderJuniorSupply, orderJuniorRedeem);
-            seniorTranche.epochUpdate(currentEpoch, 0, 0, 0, orderSeniorSupply, orderSeniorRedeem);
+            juniorTranche.epochUpdate(currentEpoch, 0, 0, 0, 0, 0);
+            seniorTranche.epochUpdate(currentEpoch, 0, 0, 0, 0, 0);
+
+            // assessor performs re-balancing according to new NAV
```
manuelpolzhofer, line 192:
```suggestion
            // assessor performs re-balancing
```

ilinzweilin (approved):

PR #496, opened 2020-12-04
feat: nav caching

manuelpolzhofer:
- refactored nav 
- added cached totalDiscount to avoid unnecessary recalculation
- fix repay bug


manuelpolzhofer:
I will close this PR because of the new nav impl #497 the caching is not required anymore

PR #497, opened 2020-12-07
feat: nav-v2

manuelpolzhofer:
**Summary**
- instead of iterating over all buckets the next nav is calculated based on the previous nav
- only all buckets between the lastNAVUpdate and `now` need to be visited
- this changes the previous runtime from `O(n)` where `n` are all buckets between `now` and the future 
- to all buckets between lastNAVUpdate and now, which is one in the most cases. (depending on minimum epoch)
- this speeds up the discounting calc to effectively `O(1)` in the most cases
- borrow and repay are updating the latestNAV

**Details**
- In the Model and an additional HackMD


manuelpolzhofer:
I opened the PR already for better discussion.

Some tests are still broken but only because of rounding issues and in some cases between one and two wei off.

src/borrower/feed/navfeed.sol
```diff
@@ -272,70 +282,69 @@ contract NAVFeed is BaseNFTFeed, Interest, Buckets, FixedPoint {
...
+        uint nLastUpdate = uniqueDayTimestamp(lastNAVUpdate);
+
+        uint totalDiscount = rmul(latestDiscount, rpow(discountRate.value, safeSub(nnow, nLastUpdate), ONE));
 
-        while(currDate != NullDate)
-        {
-            sum = safeAdd(sum, calcDiscount(buckets[currDate].value, normalizedBlockTimestamp, currDate));
-            currDate = buckets[currDate].next;
+        uint diff = 0;
+        for(uint i = nLastUpdate; i < nnow; i = i + 1 days) {
```
offerijns, line 317: I think this loop could use a comment explaining its intention, it took me a while to figure out. If I understand it correctly:

```suggestion
        // Loop over the loans which matured in between the last NAV update and now.
        // Then remove their discounted future value from the total discount as they are overdue.
        for(uint i = nLastUpdate; i < nnow; i = i + 1 days) {
```

src/borrower/feed/navfeed.sol
```diff
@@ -51,10 +58,11 @@ contract NAVFeed is BaseNFTFeed, Interest, Buckets, FixedPoint {
     // discount rate applied on every asset's fv depending on its maturityDate. The discount decreases with the maturityDate approaching.
     Fixed27 public discountRate;
 
-    // approximatedNAV is calculated in case of borrows & repayments between epoch executions.
+    // latestNAV is calculated in case of borrows & repayments between epoch executions.
     // It decreases/increases the NAV by the repaid/borrowed amount without running the NAV calculation routine.
     // This is required for more accurate Senior & JuniorAssetValue estimations between epochs
-    uint public approximatedNAV;
+    uint public latestNAV;
+    uint public latestDiscount;
```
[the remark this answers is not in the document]
manuelpolzhofer: Yes, the naming can be improved.

My reasoning behind it was the following, you are right in the moment of calculating it, the `lastest` is the `present` but if nobody interacts with the contracts and on the next day somebody only reads the value. It is not the present discount anymore only the discount from the last time we calculated it.

Mhm but you are right it maybe is a bit confusing. I will definitely add more comments

src/borrower/feed/navfeed.sol
```diff
@@ -227,25 +235,25 @@ contract NAVFeed is BaseNFTFeed, Interest, Buckets, FixedPoint {
...
     function repay(uint loan, uint amount) external auth returns (uint navDecrease) {
+        calcUpdateNAV();
         navDecrease = _repay(loan, amount);
-        if (navDecrease > approximatedNAV) {
-            approximatedNAV = 0;
-        }
 
-        if(navDecrease < approximatedNAV) {
-            approximatedNAV = safeSub(approximatedNAV, navDecrease);
+        if(navDecrease < latestNAV) {
```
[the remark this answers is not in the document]
manuelpolzhofer: good catch. 💯  I introduced the `latestDiscount` variable later.

We could only check for 
 `if(navDecrease < lastestDiscount)`

assuming `lastestNAV` is always `>= lastestDiscount`

src/borrower/feed/navfeed.sol
manuelpolzhofer: Yes,  you are correct. Technically we iterate over the buckets which are loans grouped by different maturity dates

ilinzweilin: We can get rid of the the loop & remove the overdue loans i a sep writeoff function

manuelpolzhofer:
This PR can be closed because we opened a dedicated feature branch for it.

PR #499, opened 2020-12-11
feat: mkr assessor

manuelpolzhofer:
please review #498 first
- added assessor wrapper for mkr

### DIFF
--- a/src/lender/adapters/mkr/assessor.sol
+++ b/src/lender/adapters/mkr/assessor.sol
@@ -1,0 +1,86 @@
+// Copyright (C) 2020 Centrifuge
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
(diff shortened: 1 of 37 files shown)


src/lender/adapters/mkr/assessor.sol
```diff
@@ -0,0 +1,90 @@
...
+    function depend(bytes32 contractName, address addr) public auth {
+        if (contractName == "clerk") {
+            clerk = ClerkLike(addr);
+        } else {
+            super.depend(contractName, addr);
+        }
+    }
+
+    // calculates the tokenPrice based on the nav and the reserve
+    function calcSeniorTokenPrice(uint nav_, uint) public view returns(uint) {
```
ilinzweilin, line 37: what is the name of the sec param?

ilinzweilin (approved):
👍

src/lender/adapters/mkr/assessor.sol
manuelpolzhofer: We are not using it in the wrapper otherwise it would be a warning.

PR #500, opened 2020-12-14
fix: nav repay on maturity date

manuelpolzhofer:


ilinzweilin (approved):

src/borrower/feed/navfeed.sol
```diff
@@ -255,8 +255,9 @@ contract NAVFeed is BaseNFTFeed, Interest, Buckets, FixedPoint {
         uint maturityDate_ = maturityDate[nftID_];
 
 
-        // no fv decrease calculation needed if maturaity date is in the past 
-        if (maturityDate_ < block.timestamp) {
+        // no fv decrease calculation needed if maturity date is in the past
+        if (maturityDate_ < uniqueDayTimestamp(block.timestamp)) {
+            // if a loan is overdue the portfolio value is equal to the existing debt multiplied with a write off factor
```
offerijns, line 260: In this case it's not (yet) multiplied by a write off factor, right? If I understand it correctly:
```suggestion
            // if a loan is overdue, the portfolio value is initially equal to the existing debt
            // it will be reduced by a write off factor once it is moved to a write off group
```

offerijns (approved):

src/borrower/feed/navfeed.sol
manuelpolzhofer: We assume if the loan is overdue it is already moved to the write-off.

The loan owner could still repay the complete debt, however in that case we should actually multiple the return amount first with the write-off factor to be more precise with the approximatedNAV

PR #504, opened 2020-12-17
add: mat buffer to clerk

ilinzweilin:


src/lender/adapters/mkr/clerk.sol
```diff
@@ -238,7 +246,7 @@ contract Clerk is Auth, Math {
     // returns the required security margin for the DROP tokens
     function mat() public returns (uint) {
         (, uint256 mat) = spotter.ilks(mgr.ilk());
-        return mat; //  e.g 150% denominated in RAY
+        return safeAdd(mat, matBuffer); //  e.g 150% denominated in RAY
```
manuelpolzhofer, line 249: maybe we could calculate it once instead of adding it all the time. would it be cheaper?

manuelpolzhofer (approved):
LGTM

PR #506, opened 2020-12-21
add - heal function

ilinzweilin:


src/lender/adapters/mkr/clerk.sol
```diff
@@ -224,9 +236,37 @@ contract Clerk is Auth, Math {
...
+        if (collatDeficitDAI > 0) {
+            heal(collatDeficitDAI);
+        }
+    }
+
     // checks if the Maker credit line increase could violate the pool constraints // -> make function pure and call with current pool values approxNav
-    function validate(uint juniorSupplyDAI, uint juniorRedeemDAI, uint seniorSupplyDAI, uint seniorRedeemDAI) internal {
-        require((coordinator.validate(juniorSupplyDAI, juniorRedeemDAI, seniorSupplyDAI, seniorRedeemDAI) == 0), "supply not possible, pool constraints violated");
+    function validate(uint juniorSupplyDAI, uint juniorRedeemDAI, uint seniorSupplyDAI, uint seniorRedeemDAI) internal returns (int) {
+        return coordinator.validate(juniorSupplyDAI, juniorRedeemDAI, seniorSupplyDAI, seniorRedeemDAI);
```
manuelpolzhofer, line 269: I think this has an impact on every validate call in the clerk.

Like `raise` etc.

We need to check for equal zero at least in the raise method.

We should also add a constant for that.

src/lender/adapters/mkr/clerk.sol
```diff
@@ -152,13 +162,15 @@ contract Clerk is Auth, Math {
...
-        validate(0, protectionDAI, overcollAmountDAI, 0);
+        require((validate(0, protectionDAI, overcollAmountDAI, 0) == 0), "supply not possible, pool constraints violated");
         // increase MKR crediline by amount
         creditline = safeAdd(creditline, amountDAI);
     }
 
     // mint DROP, join DROP into cdp, draw DAI and send to reserve
     function draw(uint amountDAI) public auth active {
+        // make sure ther eis no collateral deficit before drawing out new DAI
+        require(collatDeficit() == 0, "please heal cdp first");
```
manuelpolzhofer, line 173: I think we need to brainstorm about this require a bit.

Currently the coordinator considers the total creditline as available and would allow redeem on it.

In the executeEpoch transaction the reserve would fail to draw the DAI

manuelpolzhofer (approved):
LGTM

PR #510, opened 2020-12-29
feat: added creditBuffer to totalBalance in assessor + drawOnDemand tests

manuelpolzhofer:
**New Total Balance Function in Assessor**
- over time the `clerk.remainingCredit` will decrease because of the accumulated debt interest, therefore a buffer is reduced from the  remainingCredit to prevent the usage of currency which is not available.
- the buffer variable uses the stability fee multiplied with time constant( which should be way greater than minimumEpochClose time)

**Tests**
- added integration test for on demand draw in reserve

### DIFF
--- a/src/lender/adapters/mkr/assessor.sol
+++ b/src/lender/adapters/mkr/assessor.sol
@@ -104,7 +104,8 @@ contract MKRAssessor is Assessor {
         uint debt = clerk.debt();
         // over the time the remainingCredit will decrease because of the accumulated debt interest
         // therefore a buffer is reduced from the  remainingCredit to prevent the usage of currency which is not available
-        uint debtIncrease = safeSub(chargeInterest(debt, clerk.stabilityFee(), safeAdd(block.timestamp, creditBufferTime)), debt);
+        uint debtIncrease = safeSub(rmul(rpow(clerk.stabilityFee(),
+            safeSub(safeAdd(block.timestamp, creditBufferTime), block.timestamp), ONE), debt), debt);
 
         if(clerk.remainingCredit() > debtIncrease) {
             return safeAdd(reserve.totalBalance(), safeSub(clerk.remainingCredit(), debtIncrease));
--- a/src/lender/adapters/mkr/assessor.sol
+++ b/src/lender/adapters/mkr/assessor.sol
@@ -20,10 +20,22 @@ interface ClerkLike {
     function remainingCredit() external view returns (uint);
     function juniorStake() external view returns (uint);
     function remainingOvercollCredit() external view returns (uint);
+    function stabilityFee() external view returns(uint);
+    function debt() external view returns(uint);
 }
 
 contract MKRAssessor is Assessor {
     ClerkLike public clerk;
+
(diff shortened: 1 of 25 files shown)


src/lender/adapters/mkr/assessor.sol
```diff
@@ -89,7 +101,17 @@ contract MKRAssessor is Assessor {
     }
 
     function totalBalance() public view returns(uint) {
-        return safeAdd(reserve.totalBalance(), clerk.remainingCredit());
+        uint debt = clerk.debt();
+        // over the time the remainingCredit will decrease because of the accumulated debt interest
+        // therefore a buffer is reduced from the  remainingCredit to prevent the usage of currency which is not available
+        uint debtIncrease = safeSub(rmul(rpow(clerk.stabilityFee(),
+            safeSub(safeAdd(block.timestamp, creditBufferTime), block.timestamp), ONE), debt), debt);
```
ilinzweilin, line 108: `safeSub(safeAdd(block.timestamp, creditBufferTime), block.timestamp) = creditBufferTime`

src/lender/reserve.sol
```diff
@@ -130,6 +130,7 @@ contract Reserve is Math, Auth {
             if(drawAmount > left) {
                 drawAmount = left;
             }
+
```
ilinzweilin, line 133: we need a check in the  _deposit function if the epoche is in submission before we call wipe, otherwise repayments will fail. As wipe is only possible if epoch is not in submission period

src/lender/coordinator.sol
```diff
@@ -586,7 +584,7 @@ contract EpochCoordinator is Auth, Math, FixedPoint  {
         // assessor performs senior debt reBalancing according to new ratio
         assessor.changeSeniorAsset(seniorSupply, seniorRedeem);
         // the new reserve after this epoch can be used for new loans
-        assessor.changeReserveAvailable(newReserve);
+        assessor.changeBorrowAmountEpoch(newReserve);
         // reset state for next epochs
         lastEpochExecuted = epochID;
```
ilinzweilin, line 589: I think we should call wipe here, in case there are funds sitting in the reserve that were not able to transferred to the mgr due to epoch submission period

ilinzweilin (approved):

src/lender/adapters/mkr/assessor.sol
manuelpolzhofer: True. 👍 
It happened because i wanted to use the interest lib first.

src/lender/reserve.sol
manuelpolzhofer: Yes, still on my todo list. I will push it in a separated PR

src/lender/coordinator.sol
manuelpolzhofer: We should  set the flag `   submissionPeriod = false;` before we update the tranches.
Thats a way to guarantee that execute epoch can always `draw `or `wipe`

ilinzweilin (approved):

PR #513, opened 2021-01-04
feat: seniorBalance with buffer in mkrAssessor + integration test

manuelpolzhofer:
please review first #512 
- own function for remainingCredit in mkrAssessor which subtracts the buffer
   - method is used by remainingCredit and for overCollCredit
- added integration tests for clerk.heal and clerk.sink


ilinzweilin (approved):
nice

PR #516, opened 2021-01-06
fix: tranche.safePayout uses epochReserve instead of reserve.totalBalance

manuelpolzhofer:
the reserve.totalBalance is not correct form the tranche perspective because it can use the mkr creditline.


ilinzweilin (approved):
👍

PR #533, opened 2021-03-12
fix: tranche currency request can't be fulfilled

manuelpolzhofer:


src/lender/tranche.sol
```diff
@@ -276,14 +278,24 @@ contract Tranche is Math, Auth, FixedPoint {
...
+
         if(currencyAmount > max) {
+            // currently reserve can't fulfill the entire request
             currencyAmount = max;
         }
         reserve.payout(currencyAmount);
+        return currencyAmount;
     }
 
+    function balanceCurrencyRequest() public {
```
offerijns, line 292: I think the naming now is a bit confusing. Maybe we could rename `requestCurrency` to `requestedCurrency` and rename `balanceCurrencyRequest()` to `payoutRequestedCurrency()`?

offerijns (approved):
I left one comment about naming, but I think this is a great idea 👍 Ideally we would refactor the epoch execution more thoroughly to prevent this from happening, but that's a much more drastic change which would require more time and reviews. This is a much better quick fix which actually works pretty neatly!

ilinzweilin (approved):

PR #536, opened 2021-04-07
feat: add pool admin

offerijns:


src/lender/admin/pool.sol
```diff
@@ -0,0 +1,107 @@
...
+interface AssessorLike {
+    function file(bytes32 name, uint256 value) external;
+}
+
+interface LendingAdapter {
+    function raise(uint256 amount) external;
+    function sink(uint256 amount) external;
+}
+
+contract MemberlistLike {
```
lucasvo, line 28: why not `interface`?

src/lender/admin/pool.sol
```diff
@@ -0,0 +1,107 @@
...
+    function updateMember(address usr, uint256 validUntil) public;
+    function updateMembers(address[] memory users, uint256 validUntil) public;
+}
+
+// Wrapper contract for various pool management tasks.
+contract PoolAdmin is Auth {
+
```
lucasvo, line 35: I'd add events similarly to how we use them in: https://github.com/centrifuge/tinlake-maker-lib/blob/master/src/mgr.sol#L87 (especially for file)

src/lender/admin/pool.sol
```diff
@@ -0,0 +1,107 @@
...
+    // Admins can manage pools, but have to be added and can be removed by any ward on the PoolAdmin contract
+    mapping(address => uint256) public admins;
+
+    constructor() public {
+        wards[msg.sender] = 1;
+    }
+
+    function depend(bytes32 contractName, address addr) public auth {
+        if (contractName == "assessor") {
```
lucasvo, line 51: add a File() event here

src/lender/admin/pool.sol
```diff
@@ -0,0 +1,107 @@
...
+contract MemberlistLike {
+    function updateMember(address usr, uint256 validUntil) public;
+    function updateMembers(address[] memory users, uint256 validUntil) public;
+}
+
+// Wrapper contract for various pool management tasks.
+contract PoolAdmin is Auth {
+  
+    AssessorLike public assessor;
+    LendingAdapter public lending;
```
lucasvo, line 37: Should we call this maker? will we have more than once instance? What about aave? Will we need this for aave too?

offerijns: I based this on how it's called in the reserve: https://github.com/centrifuge/tinlake/blob/develop/src/lender/reserve.sol

src/lender/admin/pool.sol
offerijns: No good reason, fixed now!

src/lender/admin/pool.sol
offerijns: I'm not sure whether we're going to need it for aave, but I think keeping it a bit generic can't hurt either way. I think it's unlikely that we'll have more than one instance right? Since if we had a pool that was integrated both with Maker and Aave, we would probably create a "wrapper lending adapter" which links to the separate lending adapters.

lucasvo (approved):

PR #539, opened 2021-04-09
Rename matBuffer to mat and simplify

lucasvo:


src/lender/adapters/mkr/clerk.sol
```diff
@@ -102,33 +98,27 @@ interface ERC20Like {
...
     ERC20Like public dai;
-    ERC20Like public collateral;
 
-    // buffer to add on top of mat to avoid cdp liquidation => default 1%
-    uint matBuffer = 0.01 * 10**27;
 
     // adapter functions can only be active if the tinlake pool is currently not in epoch closing/submissions/execution state
-    modifier active() { require(activated(), "epoch-closing"); _; }
-
-    function activated() public view returns(bool) {
```
manuelpolzhofer, line 129: there is a reason why this function exists. It is needed in the reserve

PR #544, opened 2021-04-20
feat: clerk added wipeThreshold and fix sink borrowAmountEpoch

manuelpolzhofer:
### Content
   **clerk added wipeThreshold**
   - the debt is only repaid if amount is higher than the  wipe threshold
   -  repaying a lower amount would cause more cost in gas fees than the debt reduction
   
   **clerk  sink epochBorrowAmount**
   - the sink amount can be higher than assessor. borrowAmountEpoch
   - in that case  assessor. borrowAmountEpoch can be set to zero otherwise the safeSub would fail

### DIFF
--- a/src/lender/adapters/mkr/clerk.sol
+++ b/src/lender/adapters/mkr/clerk.sol
@@ -122,6 +122,8 @@ contract Clerk is Auth, Interest {
     ERC20Like public dai;
     ERC20Like public collateral;
 
+    uint public constant WAD = 10*18;
+
     // buffer to add on top of mat to avoid cdp liquidation => default 1%
     uint public matBuffer = 0.01 * 10**27;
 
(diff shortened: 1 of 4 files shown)


src/lender/adapters/mkr/clerk.sol
```diff
@@ -128,6 +128,10 @@ contract Clerk is Auth, Interest {
     // collateral tolerance accepted because of potential rounding problems
     uint public collateralTolerance = 10;
 
+    // the debt is only repaid if amount is higher than the threshold
+    // repaying a lower amount would cause more cost in gas fees than the debt reduction
+    uint public wipeThreshold = 10**18;
```
offerijns, line 133: Maybe we should define `uint WAD = 10**18`?

offerijns (approved):

src/lender/adapters/mkr/clerk.sol
```diff
@@ -122,12 +122,18 @@ contract Clerk is Auth, Interest {
     ERC20Like public dai;
     ERC20Like public collateral;
 
+    uint public constant WAD = 10*18;
+
     // buffer to add on top of mat to avoid cdp liquidation => default 1%
     uint public matBuffer = 0.01 * 10**27;
```
offerijns, line 128: This can actually be changed like so, right? Since the contract inherits from the Interest contract, which inherits from the Math contract, which contains this variable.
```suggestion
    uint public matBuffer = 0.01 * ONE;
```

ilinzweilin (approved):

PR #555, opened 2021-04-29
fix: transferFrom with transfer in all erc20 interactions

manuelpolzhofer:


src/lender/reserve.sol
```diff
@@ -54,6 +54,7 @@ contract Reserve is Math, Auth {
         wards[msg.sender] = 1;
         currency = ERC20Like(currency_);
         pot = address(this);
+        currency.approve(pot, uint(-1));
```
offerijns, line 57: Although it seems unlikely that we'll ever use this, I think this should also be added in two places in depend(), both when depending the pot and the currency, right?

offerijns (approved):
One minor comment, looks great!

src/lender/reserve.sol
manuelpolzhofer: If we change the pot (which means unequal `address(this)`) than this contract can't give the approval.

The approval would need to happen outside of this contract.

If we change the currency and the pot is still `address(this)` than we you are right and we could add it in the depend.

PR #559, opened 2021-05-14
feat: make mkr assessor the default assessor

offerijns:


src/lender/assessor.sol
```diff
@@ -276,4 +334,31 @@ contract Assessor is Definitions, Auth, Interest {
 
         return safeSub(ONE, rdiv(seniorAsset, assets));
     }
+
+    // returns the remainingCredit plus a buffer for the interest increase
+    function remainingCredit() public view returns(uint) {
+        if (address(lending) == address(0)) {
```
manuelpolzhofer, line 340: I think we could make a modifier for all lender related methods.

something like:

```javascript
   modifier lenderActive() {
      if (address(lending) != address(0)) {
         _;
      }
      return 0;

   }
```

We would need to check if this works.

src/lender/assessor.sol
```diff
@@ -164,19 +193,30 @@ contract Assessor is Definitions, Auth, Interest {
...
         uint seniorAssetValue = calcExpectedSeniorAsset(seniorDebt(), seniorBalance_);
 
         if(totalAssets < seniorAssetValue) {
             return 0;
         }
 
-        return rdiv(safeSub(totalAssets, seniorAssetValue), juniorTranche.tokenSupply());
+        // the junior tranche only needs to pay for the mkr over-collateralization if
+        // the mkr vault is liquidated, if that is true juniorStake=0
+        uint juniorStake = 0;
```
manuelpolzhofer, line 213: We could make a method for getting the juniorStake together with the modifier which checks if a lender exists otherwise return 0.

src/lender/assessor.sol
```diff
@@ -239,14 +279,32 @@ contract Assessor is Definitions, Auth, Interest {
...
+
+    // returns the current NAV
+    function currentNAV() public view returns(uint) {
+        return navFeed.currentNAV();
+    }
+
+    // returns the approximated NAV for gas-performance reasons
+    function getNAV() public view returns(uint) {
+        return navFeed.approximatedNAV();
+    }
```
manuelpolzhofer, line 306: We can refactor this methods after the more efficient NAV PR is merged.

manuelpolzhofer (approved):
LGTM.

I have one small idea to maybe use a `modifier` for the lender zero checks

I think we should check if we have a good test coverage for the case if we use a lending adapter with no over-collateralization.

offerijns:
> I think we should check if we have a good test coverage for the case if we use a lending adapter with no over-collateralization.

Added https://github.com/centrifuge/tinlake/issues/560 for this.

manuelpolzhofer (approved):

PR #564, opened 2021-05-20
refactor: upgrade to solidity 0.6.12

offerijns:
- set updated version
- add license headers
- add virtual/override modifiers
- add view/pure modifiers 
- upgrade dependencies, add `ds-note` as direct dependency
- make some contracts abstract


src/lender/adapters/mkr/deployer.sol
```diff
@@ -58,7 +44,7 @@ contract MKRLenderDeployer is LenderDeployer {
         mkrDeployer = address(1);
     }
 
-    function deploy() public {
+    function deployMkr() public {
```
manuelpolzhofer, line 47: Was this because of overriding?

src/lender/coordinator.sol
```diff
@@ -30,30 +17,30 @@ interface ReserveLike {
     function balance() external;
 }
 
-contract AssessorLike is FixedPoint {
+abstract contract AssessorLike is FixedPoint {
```
manuelpolzhofer, line 20: I think here it is fine to still use an abstract contract otherwise it might be tricky with the Fixed27 type.

It is easier to read if we can use them in the `Like` definitions.

src/test/system/lender/lender_scenarios.t.sol
```diff
@@ -506,9 +492,7 @@ contract LenderSystemTest is TestSuite, Interest {
 
         uint preBalance = currency.balanceOf(seniorInvestor_);
 
-        (uint seniorRedeem,,,) = coordinator.order();
```
manuelpolzhofer, line 509: Was this variable not used?

test.sh
```diff
@@ -0,0 +1,12 @@
+RESULT=$(dapp test)
```
manuelpolzhofer, line 1: 💯

src/lender/adapters/mkr/test/clerk.t.sol
```diff
@@ -663,7 +642,6 @@ contract ClerkTest is Assertions, Interest {
         assertEq(lockedCollateralDAI, 110 ether);
         assertEq(requiredCollateralDAI, 115 ether);
         // partial healing
-        uint healingAmount = safeDiv(safeSub(requiredCollateralDAI, lockedCollateralDAI), 2); // healing amount = 2
         assessor.setReturn("balance", 200 ether);
         heal(10, 4, false);
```
manuelpolzhofer, line 646: This line seems a bit strange @ilinzweilin

manuelpolzhofer (approved):
LGTM.

thanks for that @offerijns

src/lender/token/erc20.sol
```diff
@@ -0,0 +1,122 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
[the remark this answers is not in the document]
offerijns: Yes exactly!

src/test/system/lender/lender_scenarios.t.sol
offerijns: Nope!

src/borrower/shelf.sol
```diff
@@ -1,52 +1,31 @@
...
-    function transfer(address, uint) public returns (bool);
-    function approve(address, uint) public;
+    function balanceOf(address) public virtual view returns (uint);
+    function transferFrom(address,address,uint) public virtual returns (bool);
+    function transfer(address, uint) public virtual returns (bool);
+    function approve(address, uint) public virtual;
 }
 
-contract PileLike {
+abstract contract PileLike {
```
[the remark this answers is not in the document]
offerijns: Good point, changed.

src/lender/adapters/mkr/deployer.sol
offerijns: It wasn't actually necessary, thought it was at first. Fixed now.

src/lender/admin/member.sol
```diff
@@ -30,10 +17,20 @@ contract MemberAdmin is Auth {
...
     mapping (address => uint) public admins;
 
+    event RelyAdmin(address indexed usr);
+    event DenyAdmin(address indexed usr);
+
     modifier admin { require(admins[msg.sender] == 1); _; }
 
-    function relyAdmin(address usr) public auth note { admins[usr] = 1; }
-    function denyAdmin(address usr) public auth note { admins[usr] = 0; }
+    function relyAdmin(address usr) public auth {
```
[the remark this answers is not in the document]
offerijns: No, the problem was that `note` was previously included through the `tinlake-auth` dependency, but `ds-note` was actually removed from `tinlake-auth` now. So rather than change this to include `ds-note` here now directly, I thought it was cleaner to then just rewrite it properly.

PR #577, opened 2021-06-08
refactor adapter deployer

ilinzweilin:


src/lender/adapters/deployer.sol
```diff
@@ -60,59 +60,33 @@ contract AdapterDeployer {
...
-
-    function deploy(bool wireMgr) public {
-		setupMkrAdapter(wireMgr);
-    }
-
-    function deploy() public {
-        deploy(false);
-    }
+    function deployClerk() public {
+        require(lenderDeployer.seniorToken() != address(0));
```
offerijns, line 72:
```suggestion
        require(lenderDeployer.clerk() != address(0));
```

offerijns (changes requested):

src/lender/adapters/deployer.sol
ilinzweilin: I guess you mean `require(lenderDeployer.clerk() == address(0));`

ilinzweilin: seniorToken check can stay, as it makes sure that the contract has to be deployed prior to deploying the clerk

offerijns (approved):
LGTM

PR #578, opened 2021-06-14
chore: NAV finds first relevant bucket with a pointer

manuelpolzhofer:
- instead of finding the first relevant bucket for the currentDiscount by increasing the current block.timestamp by a day
- a pointer keeps track of the `first bucket` in the future
   - in the next call the this day might be already in the past but we start from here to find the first relevant bucket

the proposed solution is more efficient with maturity dates far in the future

### DIFF
--- a/src/borrower/feed/navfeed.sol
+++ b/src/borrower/feed/navfeed.sol
@@ -299,7 +299,7 @@ contract NAVFeed is BaseNFTFeed, Interest, Buckets, FixedPoint {
         }
 
         // next time we can start here to find the next first bucket in the future
-        discountStartPointer = currDate;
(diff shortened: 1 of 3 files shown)


src/borrower/feed/navfeed.sol
```diff
@@ -330,4 +346,23 @@ contract NAVFeed is BaseNFTFeed, Interest, Buckets, FixedPoint {
     function dateBucket(uint timestamp) public view returns (uint) {
         return buckets[timestamp].value;
     }
+
+    function addBucket(uint timestamp, uint value) public override {
```
ilinzweilin, line 350: why is this function public

ilinzweilin (approved):

PR #584, opened 2021-07-27
feat: NAV v2

offerijns:


src/borrower/feed/navfeed.sol
```diff
@@ -189,22 +190,46 @@ abstract contract NAVFeed is Auth, Discounting {
...
+            }
+        }
+
         pile.changeRate(loan, WRITEOFF_RATE_GROUP_START + writeOffGroupIndex_);
     }
 
     // --- NAV calculation ---
     function currentNAV() public view returns(uint) {
-        return safeAdd(currentDiscount(), currentWriteOffs());
+        return safeAdd(safeAdd(currentDiscount(), overdueButNotWrittenOff), currentWriteOffs());
```
manuelpolzhofer, line 232: This would break our current pattern, if I understand correctly.
Currently the view method grantees a up to date NAV.

I think we should calculate the new `overdueButNotWrittenOff` value.

src/borrower/feed/navfeed.sol
```diff
@@ -217,23 +242,15 @@ abstract contract NAVFeed is Auth, Discounting {
 
         uint totalDiscount = rmul(latestDiscount, rpow(discountRate.value, safeSub(nnow, nLastUpdate), ONE));
 
-        // Loop over the loans which matured in between the last NAV update and now.
-        // Then remove their discounted future value from the total discount as they are overdue.
-        uint diff = 0;
-        for(uint i = nLastUpdate; i < nnow; i = i + 1 days) {
```
manuelpolzhofer: We still need to calculate the error and subtract it. 
I think here we could also calculate the new `overdueButNotWrittenOff`  and add it again, because we anyway need to iterate over the buckets from the lastUpdate until now.

src/borrower/feed/navfeed.sol
```diff
@@ -243,8 +260,21 @@ abstract contract NAVFeed is Auth, Discounting {
...
     function calcUpdateNAV() public returns(uint) {
+        uint nnow = uniqueDayTimestamp(block.timestamp);
+        uint nLastUpdate = uniqueDayTimestamp(lastNAVUpdate);
+
+        if (nLastUpdate < nnow) {
+            // Loop over the loans which matured in between the last NAV update and now, and save their future value.
+            uint overdueButNotWrittenOff_ = 0;
+            for (uint i = nLastUpdate; i < nnow; i = i + 1 days) {
+                overdueButNotWrittenOff_ = safeAdd(overdueButNotWrittenOff_, buckets[i]);
+            }
```
manuelpolzhofer, line 271: 👍

src/borrower/feed/navfeed.sol
```diff
@@ -3,267 +3,155 @@ pragma solidity >=0.6.12;
...
+    mapping (uint => uint)      public ceilingRatio;    // risk => ceilingRatio
+    mapping (bytes32 => uint)   public maturityDate;    // nftID => maturityDate
+    mapping (uint => uint)      public thresholdRatio;  // risk => thresholdRatio
+    mapping (uint => Fixed27)   public recoveryRatePD;  // risk => recoveryRatePD
+    mapping (bytes32 => uint)   public risk;            // nftID => risk
+    mapping (bytes32 => uint)   public nftValues;       // nftID => nftValues
+    mapping (bytes32 => uint)   public futureValue;     // nftID => futureValue
+    mapping (uint => uint)      public borrowed;        // loan => borrowed
+    mapping (uint => bool)      public writeOffOverride;// loan => writeOffOverride
+    mapping (uint => uint)      public buckets;         // timestamp => bucket
```
manuelpolzhofer, line 45: I think we should look into optimisation here. 
If we use a struct, we can use uint128 and uint64 etc.

Would be nice to compare the gas differences etc.

src/borrower/feed/navfeed.sol
```diff
@@ -273,106 +161,272 @@ contract NAVFeed is BaseNFTFeed, Interest, Buckets, FixedPoint {
...
-        uint normalizedBlockTimestamp = uniqueDayTimestamp(block.timestamp);
-        uint sum = 0;
+    // --- NAV calculation ---
+    function currentNAV() public view returns(uint) {
+        uint nnow = uniqueDayTimestamp(block.timestamp);
+        uint nLastUpdate = uniqueDayTimestamp(lastNAVUpdate);
+        uint overdueButNotWrittenOff_ = 0;
+        for (uint i = nLastUpdate; i < nnow; i = i + 1 days) {
+            overdueButNotWrittenOff_ = safeAdd(overdueButNotWrittenOff_, buckets[i]);
+        }
```
manuelpolzhofer, line 238: It would be nice to calculate the `overdueButNotWrittenOff_` in the `currentDiscount` loop.
I think this would be cheaper because the additional `i` increment would not be required

src/borrower/feed/navfeed.sol
```diff
@@ -273,106 +161,272 @@ contract NAVFeed is BaseNFTFeed, Interest, Buckets, FixedPoint {
...
 
-    // returns the NAV (net asset value) of the pool
-    function _currentNAV() internal view returns(uint, uint) {
-        // calculates the NAV for ongoing loans with a maturityDate date in the future
-        (uint nav_, uint nextStartPointer) = _calcTotalDiscount();
-        // include ovedue assets to the current NAV calculation
-        for (uint i = 0; i < writeOffs.length; i++) {
+    function currentWriteOffs() public view returns(uint) {
+        uint sum = 0;
+        for (uint i = 0; i < writeOffGroups.length; i++) {
```
manuelpolzhofer, line 272: I think we could safe some gas if the loop would start at `WRITEOFF_RATE_GROUP_START`

src/borrower/feed/principal.sol
```diff
@@ -0,0 +1,89 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity >=0.6.12;
+pragma experimental ABIEncoderV2;
+
+import "./navfeed.sol";
+
+contract PrincipalNAVFeed is NAVFeed {
```
manuelpolzhofer, line 7: nice pattern 👍

src/borrower/feed/navfeed.sol
```diff
@@ -273,106 +161,272 @@ contract NAVFeed is BaseNFTFeed, Interest, Buckets, FixedPoint {
...
     }
 
-    function calcTotalDiscount() public view returns(uint) {
-        (uint sum, ) = _calcTotalDiscount();
-        return sum;
+    function _writeOff(uint loan, uint writeOffGroupIndex_, bytes32 nftID_, uint maturityDate_) internal {
+        if (pile.loanRates(loan) < WRITEOFF_RATE_GROUP_START) {
+            // this is the first writeoff, remove future value for loan from bucket
+            uint fv = futureValue[nftID_];
+            buckets[maturityDate_] = safeSub(buckets[maturityDate_], fv);
```
manuelpolzhofer, line 218: This would be only required if the loan is written-off before the maturity date. 

We could safe some gas here.

In that case I think we also need to substract the current PV from the stored totalDiscount

src/borrower/feed/navfeed.sol
```diff
@@ -273,106 +161,272 @@ contract NAVFeed is BaseNFTFeed, Interest, Buckets, FixedPoint {
...
+
+            if (uniqueDayTimestamp(lastNAVUpdate) >= maturityDate_) {
+                overdueButNotWrittenOff = secureSub(overdueButNotWrittenOff, fv);
+            }
+
+            latestDiscount = secureSub(latestDiscount, fv);
```
manuelpolzhofer, line 224: I think this is not  always correct, if the write-off happens after the maturity date and `lastNAVUpdate > maturityDate`.

In that case the PV of the loan is already in `overdueButNotWrittenOff` and not in `latestDiscount` anymore.

manuelpolzhofer: If the write-off happens before the maturity date, we need to substract

```javascript
  uint pv = safeAdd(diff, rmul(fv, rpow(discountRate.value, safeSub(nnow, i), ONE)));
```

src/borrower/feed/navfeed.sol
```diff
@@ -273,106 +161,272 @@ contract NAVFeed is BaseNFTFeed, Interest, Buckets, FixedPoint {
...
-    function calcTotalDiscount() public view returns(uint) {
-        (uint sum, ) = _calcTotalDiscount();
-        return sum;
+    function _writeOff(uint loan, uint writeOffGroupIndex_, bytes32 nftID_, uint maturityDate_) internal {
+        if (pile.loanRates(loan) < WRITEOFF_RATE_GROUP_START) {
+            // this is the first writeoff, remove future value for loan from bucket
+            uint fv = futureValue[nftID_];
+            buckets[maturityDate_] = safeSub(buckets[maturityDate_], fv);
+
+            if (uniqueDayTimestamp(lastNAVUpdate) >= maturityDate_) {
```
manuelpolzhofer, line 220: I think it might be easier if we require an up to date NAV instead of handling the edge cases.

`calcUpdateNAV` should anyway be much cheaper now.

src/borrower/feed/test/navfeed.t.sol
```diff
@@ -336,15 +309,29 @@ contract NAVTest is DSTest, Math {
...
-    function testChangeMaturityDebtRepaid() public {
-        // test uses loan 2
-        testRepay();
-        bytes32 nftID = feed.nftID(2);
-        // no loan debt exists anymore -> maturity date change possible
-        uint newDate = feed.maturityDate(nftID) + 2 days;
-        // no loan debt exists -> maturity date change possible
-        feed.file("maturityDate", nftID, newDate);
-        assertEq(feed.maturityDate(nftID), feed.uniqueDayTimestamp(newDate));
+    function testOverdueLoan() public {
```
manuelpolzhofer, line 312: 👍

manuelpolzhofer (commented):
nice work so far. Assuming this is still WIP.

I added some ideas for gas optimisation and some questions about potential problems in the write off function

src/borrower/feed/navfeed.sol
manuelpolzhofer: Maybe we can rename the  totalDiscount function into presentValue and it returns the totalDiscount and an up to date `overdueButNotWrittenOff` or the totalDiscount + overdueButNotWrittenOff

src/borrower/feed/navfeed.sol
offerijns: The writeoff groups don't start at `WRITEOFF_RATE_GROUP_START`. Only the rate groups in the pile that correspond to these writeoff groups start at `WRITEOFF_RATE_GROUP_START`.

src/borrower/feed/navfeed.sol
offerijns: Would be nice, but I think we should look into this after we have all the functionality of NAV v2 working, including proper tests etc.

src/borrower/feed/navfeed.sol
offerijns: `WRITEOFF_RATE_GROUP_START` needs a better name :)

manuelpolzhofer: Ahh, I missed the `writeOffGroups[i].percentage.value` part.

src/borrower/feed/navfeed.sol
```diff
@@ -273,106 +161,288 @@ contract NAVFeed is BaseNFTFeed, Interest, Buckets, FixedPoint {
...
+        _writeOff(loan, writeOffGroupIndex_, nftID_, maturityDate_);
+        emit WriteOff(loan, writeOffGroupIndex_, true);
     }
 
-    function calcTotalDiscount() public view returns(uint) {
-        (uint sum, ) = _calcTotalDiscount();
-        return sum;
+    function _writeOff(uint loan, uint writeOffGroupIndex_, bytes32 nftID_, uint maturityDate_) internal {
+        // Ensure we have an up to date NAV
+        calcUpdateNAV();
```
manuelpolzhofer, line 211:
```suggestion
        if(block.timestamp > lastNAVUpdate) {
        calcUpdateNAV();
        }
```
maybe only if it is not up to date

offerijns: Good point!

PR #586, opened 2021-08-09
feat: add lock and repay events, add amounts

offerijns:


src/borrower/shelf.sol
```diff
@@ -157,9 +159,11 @@ contract Shelf is Auth, TitleOwned, Math {
     // a max ceiling needs to be defined by an oracle
     function borrow(uint loan, uint currencyAmount) external owner(loan) {
         require(nftLocked(loan), "nft-not-locked");
+
         if(address(subscriber) != address(0)) {
-            subscriber.borrowEvent(loan);
+            subscriber.borrowEvent(loan, currencyAmount);
```
manuelpolzhofer, line 164: Currently the only reason why we are doing this is for consistency?

manuelpolzhofer (approved):
LGTM

src/borrower/shelf.sol
offerijns: Yeah, this one is purely for consistency reasons. Wdyt, is there any downside to doing it this way?

PR #601, opened 2021-08-30
feat: pool admin v2

offerijns:
- divides the permissions in level 1 (liquidity management), level 2 (risk management, and level 3 (pool governance)
- moves the existing pool admin functionality into level 1
- adds methods for filing risk & writeoff groups, overriding writeoffs, and configuring risk params to level 2
- adds methods for filing most pool settings and closing/unclosing pools to level 3


src/lender/admin/pool.sol
```diff
@@ -11,31 +9,59 @@ interface LendingAdapterLike {
...
+    mapping (address => uint256) public level3_admins;
 
-    // Events
-    event Depend(bytes32 indexed contractname, address addr);
-    event File(bytes32 indexed what, bool indexed data);
-    event RelyAdmin(address indexed usr);
-    event DenyAdmin(address indexed usr);
+    modifier level1     { require(level1_admins[msg.sender] == 1 && live); _; }
+    modifier level2     { require(level2_admins[msg.sender] == 1 && live); _; }
+    modifier level3     { require(level3_admins[msg.sender] == 1 && live); _; }
```
manuelpolzhofer, line 54: Currently a level3 admin could not call a level1 method. (If the same address is not added again for level1).

Is that the intention? If yes, maybe the `level` naming is a bit missleading.

However, we could also have just one mapping called `admin_level` with values from 1 to 3.  

The modifier would check for `<=`

src/lender/admin/pool.sol
```diff
@@ -45,92 +71,231 @@ contract PoolAdmin is Auth {
...
+    event OverrideWriteOff(uint loan, uint writeOffGroupIndex);
+    event AddRiskGroup(uint risk_, uint thresholdRatio_, uint ceilingRatio_, uint rate_, uint recoveryRatePD_);
+    event AddRiskGroups(uint[] risks_, uint[] thresholdRatios_, uint[] ceilingRatios_, uint[] rates_);
+    event AddWriteOffGroup(uint rate_, uint writeOffPercentage_, uint overdueDays_);
+    event SetMatBuffer(uint value);
+    event UpdateNFTValue(bytes32 nftID_, uint value);
+    event UpdateNFTValueRisk(bytes32 nftID_, uint value, uint risk_);
+    event UpdateNFTMaturityDate(bytes32 nftID_, uint maturityDate_);
+
+    function overrideWriteOff(uint loan, uint writeOffGroupIndex_) public level2 {
```
manuelpolzhofer, line 137: One other idea would be that the permissions of a function could be changed.
```javascript
function overrideWriteOff(uint loan, uint writeOffGroupIndex_) public permission("overrideWriteOff")
```

```javascript
modifier permission(bytes32 funcName) { require(admin_level[msg.sender] >= funcAccess[funcName] && live); _; }
```

manuelpolzhofer: Could we run into such a situation in which this could be a nice feature?

src/lender/admin/pool.sol
```diff
@@ -45,92 +71,231 @@ contract PoolAdmin is Auth {
...
+
+    function relyLevel1(address usr) public level2 {
+        level1_admins[usr] = 1;
+        emit Rely(1, usr);
+    }
+
+    function denyLevel1(address usr) public level2 {
+        level1_admins[usr] = 0;
+        emit Deny(1, usr);
+    }
```
manuelpolzhofer, line 194: Maybe we can move them next to `relyLevel2` and `relyLevel3`?

src/lender/admin/pool.sol
manuelpolzhofer:
```javascript
// level 1
function name(...) public admin_level(1) 

// level 2
function name(...) public admin_level(2) 
```

src/lender/admin/pool.sol
```diff
@@ -45,92 +71,231 @@ contract PoolAdmin is Auth {
...
+
+    function addRiskGroups(uint[] memory risks_, uint[] memory thresholdRatios_, uint[] memory ceilingRatios_, uint[] memory rates_, uint[] memory recoveryRatePDs_) public level2 {
+        require(risks_.length == thresholdRatios_.length && thresholdRatios_.length == ceilingRatios_.length && ceilingRatios_.length == rates_.length, "non-matching-arguments");
+        for (uint i = 0; i < risks_.length; i++) {
+            addRiskGroup(risks_[i], thresholdRatios_[i], ceilingRatios_[i], rates_[i], recoveryRatePDs_[i]);
+        }
+    }
+
+    function addWriteOffGroup(uint rate_, uint writeOffPercentage_, uint overdueDays_) public level2 {
+        navFeed.file("writeOffGroup", rate_, writeOffPercentage_, overdueDays_);
```
manuelpolzhofer, line 155: Did you double check the order of the parameters, if they are all uint like here?

manuelpolzhofer (commented):
Overall looks good. 

I added some alternative ideas how to implement it.
I still want to double check the correct order of parameters.

src/lender/admin/pool.sol
manuelpolzhofer: We could also keep the existing three different modifier but just with one mapping

src/lender/admin/pool.sol
```diff
@@ -256,24 +243,15 @@ contract PoolAdmin {
...
-    function denyLevel2(address usr) public level3 {
-        level2_admins[usr] = 0;
-        emit Deny(2, usr);
-    }
-
-    function relyLevel3(address usr) public level3 {
-        level3_admins[usr] = 1;
-        emit Rely(3, usr);
+    function rely(address usr, uint level) public level3 {
+        require(level == 1 || level == 2 || level == 3, "invalid-level");
```
manuelpolzhofer, line 247: Why not
`require(level  <= LEVEL_3 && level > 0);

manuelpolzhofer (approved):
LGTM

PR #616, opened 2022-11-03
poolValue nav

ilinzweilin:
- NAV value = total outstanding debt
- WritingOff loans by moving them to rateGroup 1000
- Ceiling = Creditline implementation


src/borrower/feed/poolValue.sol
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    // nft => details
+    mapping (bytes32 => NFTDetails) public details;
+    // loan => details
+    mapping(uint => LoanDetails) public loanDetails;
+
+    uint public latestNAV;
+
+    uint WRITEOFF_RATE_GROUP = 1000;
```
xmxanuel, line 44:
```suggestion
    uint public constant WRITEOFF_RATE_GROUP = 1000;
```

src/borrower/feed/navfeedPV.sol
```diff
@@ -0,0 +1,222 @@
...
+
+    function isLoanWrittenOff(uint loan) public view returns(bool) {
+        return pile.loanRates(loan) == WRITEOFF_RATE_GROUP;
+    }
+
+    // --- NAV calculation ---
+    function currentNAV() public view returns(uint) {
+        uint totalDebt;
+        // calculate total debt
+        for (uint loanId = 1; loanId <= shelf.loanCount(); loanId++) {
```
xmxanuel, line 149:
```suggestion
        for (uint loanId = 1; loanId < shelf.loanCount(); loanId++) {
```
If you iterate over the loans. I think it should just be `<`. 

```solidity
    function loanCount() public view returns (uint256 totalNumber) {
        return title.count();
    }
```

**Example**.

Here is an example to illustrate.
`Title.count` starts with 1. (zero loans in the system.)

After the first loan (id = 1) `count` will be `2`.

I think this is actually a weird implementation. I don't remember why Lucas back then choose to start with loan id =1.

src/borrower/feed/navfeedPV.sol
```diff
@@ -0,0 +1,222 @@
...
+            totalDebt = safeAdd(totalDebt, pile.debt(loanId));
+        }
+
+        // substract writtenoff loans -> all writtenOff loans are moved to writeOffRateGroup
+        totalDebt = safeSub(totalDebt, pile.rateDebt(WRITEOFF_RATE_GROUP));
```
xmxanuel, line 154: If you iterate over all loan debt it make sense to remove it from the total debt.

However, the write-offs should still be added to the currentNAV in my opinion.
```
Example:
two loans:
1. debt: 100 DAI, riskGroup=1
2. debt: 80 DAI, writeOffGroup; writeOffFactor= 0.5 
```

For me, it makes sense to remove 80 DAI but you still need to add 80 DAI * 0.5 = 40 DAI to the NAV?

It would only make sense if the writeOff is always 100%. Is this the case?

src/borrower/feed/navfeedPV.sol
```diff
@@ -0,0 +1,222 @@
...
+    // returns the nftID for the underlying collateral nft
+    function nftID(uint loan) public view returns (bytes32) {
+        (address registry, uint tokenId) = shelf.shelf(loan);
+        return nftID(registry, tokenId);
+    }
+
+    // returns true if the present value of a loan is zero
+    // true if all debt is repaid or debt is 100% written-off
+    function zeroPV(uint loan) public view returns (bool) {
+        return ((pile.debt(loan) == 0) || (pile.loanRates(loan) == WRITEOFF_RATE_GROUP));
```
xmxanuel, line 215: :+1:

src/borrower/feed/navfeedPV.sol
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    // normalizes a timestamp to round down to the nearest midnight (UTC)
+    function uniqueDayTimestamp(uint timestamp) public pure returns (uint) {
```
xmxanuel, line 219: I think if the NAV is calcuated based on the `debt` which changes per second. It might make sense to remove the `uniqueDayTimestamp`  which basically enforces the NAV to be a step function per day. 

The NAV debt would increase only by day. Is this needed in this scenario for this version?

xmxanuel (commented):
I added some comments. Overall it looks good. I can follow the reasoning behind the changes in `update` and `currentNAV` in general. 

However, I still have some questions about the requirements. 

Together, with some small fixes.

src/borrower/feed/navfeedPV.sol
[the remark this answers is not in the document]
ilinzweilin: no, as the pools will only have less than 5 loans

src/borrower/feed/test/navfeed.t.sol
```diff
@@ -61,7 +61,7 @@ contract NAVTest is DSTest, Math {
         bytes32 nftID = feed.nftID(mockNFTRegistry, tokenId);
         feed.update(nftID, nftValue, risk);
         shelf.setReturn("shelf",mockNFTRegistry, tokenId);
-        pile.setReturn("debt_loan", 0);
+        pile.incDebt(tokenId, 0);
```
xmxanuel, line 64: Okay, here you are using the real pile instead of a mocked version. Correct?

src/borrower/feed/test/navfeedPV.t.sol
```diff
@@ -0,0 +1,335 @@
...
+        uint riskGroup = 1;
+        uint loan = 1;
+        uint value = 100 ether;
+        uint debt = 71 ether;
+        uint maxCeiling = 70 ether;
+        feed.update(nftID, value, riskGroup);
+
+        shelf.setReturn("shelf", address(1), 1);
+        assertEq(feed.ceiling(loan), maxCeiling);
+        pile.incDebt(loan, debt);
```
xmxanuel, line 92: Okay, to summarize. The actual `ceiling` check happens in the borrow in the nav feed which is triggered by the `borrower` subscriber event in the shelf.

We need to ensure that the subscriber is always set on the shelf.

src/test/system/scenarios.t.sol
```diff
@@ -1,179 +0,0 @@
-// SPDX-License-Identifier: AGPL-3.0-only
```
xmxanuel, line 1: Why are these secenario tests deleted?

src/test/system/borrower/repay/creditline_repay.t.sol
```diff
@@ -78,15 +93,40 @@ contract PrincipalRepayTest is BaseSystemTest {
...
+        assertEq(nftFeed.ceiling(loanId), 100 ether);
+    } 
+
     function testRepayFullDebt() public {
         uint nftPrice = 200 ether; // -> ceiling 100 ether
         uint riskGroup = 1; // -> 12% per year
 
         // expected debt after 1 year of compounding
-        uint expectedDebt = 112 ether;
+        uint expectedDebt = 112000000000000000001; // 112 ETH
```
xmxanuel, line 129:
```suggestion
        uint expectedDebt = (112 ether) + 1;
```

xmxanuel: Why are we here 1 Wei off?

src/borrower/feed/test/navfeed.t.sol
xmxanuel: Is there any specific reason?

xmxanuel (commented):
Here are my first notes on the PR. I want to investigate some of the 1 Wei rounding errors related to the debt.

src/borrower/feed/test/navfeed.t.sol
ilinzweilin: I am using the mock. The mock also has the incDebt function to keep track of multiple loan debts

src/borrower/feed/test/navfeedPV.t.sol
ilinzweilin: yes, the checks happens in the NAV. If the address wouldn't be set, the check would revert anyway, right

src/test/system/borrower/repay/creditline_repay.t.sol
ilinzweilin: our usual rounding errors, that's why the pre/post condition assertions have the 1 wei tolerance

PR #618, opened 2022-11-09
Set up deploy script

offerijns:
Sets up a deployment script using Foundry Scripting that replicates the functionality of https://github.com/centrifuge/tinlake-deploy. Fabs are deployed using the standard create2 factory. Contract addresses are logged during the deployment.

The parameters for the deployment need to be set up in a `.env` file. An example one is provided and linked from the README. The environment checking script from https://github.com/centrifuge/tinlake-deploy is added as a separate bin script.


AStox (approved):
Looks good!

PR #619, opened 2022-11-09
fix: cleanup shelf contract unreachable internal function

xmxanuel:
I started to look at the Tinlake codebase again to refresh everything. 

I noticed an internal function that is not called anymore. 
This most likely happened because of a refactoring.


offerijns (approved):
Nice find 👍

xmxanuel:
@ilinzweilin lets make two reviewers for this PR. Since it touches the core logic.

offerijns (dismissed):

AStox (approved):

offerijns (approved):

PR #627, opened 2022-11-24
Pool value nav deploy

offerijns:
- Fix deploy scripts for different NAV feeds
- Add RPC tests
- Separate CI steps for tests and format checks


script/deploy.s.sol
```diff
@@ -47,14 +47,26 @@ contract DeployScript is Script {
     }
 
     function deployBorrower(TinlakeRoot root) internal returns (address) {
+        address feedFab = address(0);
+        if (keccak256(abi.encodePacked(vm.envString("NAV_IMPLEMENTATION"))) == keccak256("PV")) {
+            feedFab = getOrDeployFab("navfeedPV.sol:NAVFeedPVFab");
```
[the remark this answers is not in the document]
xmxanuel: This was required for me to run the deploy script locally for me.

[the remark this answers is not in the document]
xmxanuel: In general I think it would make sense to use individual salts or one global salt for the Tinlake deployment.

What is the meaning  of `bytes32 salt = 0x00000000000000000000000000000000000000008b99e5a778edb02572010000;`

Maybe you do something like `bytes32 salt = keccak(abi.encoode("Tinlake")`

xmxanuel:
Hey,

some of my comments are addressing more the deploy script in general as well. 

I think it could also make sense to  use the `findCreate2Address` function instead of doing the same operations in the script.
It would mean it is one `CALL` more but it might be okay to reduce the complexity.

xmxanuel:
It might be even better to use `findCreate2AddressViaHash`

src/borrower/deployer.sol
```diff
@@ -103,7 +103,7 @@ contract BorrowerDeployer is FixedPoint {
 
     /// @notice deploys the borrower contracts and wires them together
     /// @param initNAVFeed boolean flag if a NAV feed should be deployed
-    function deploy(bool initNAVFeed) public {
+    function deploy(bool initNAVFeed, bool fileDiscountRate) public {
```
xmxanuel, line 106: Okay, so basically in the PV feed case we pass `ONE` to the borrower deployer as default `discountRate` but we never set it. 

I think this might be okay for now. In the ideal case we would have separate Borrower deployers.

README.md
```diff
@@ -53,6 +53,11 @@ When the deployment is complete, you can print the list of deployed contracts:
 ROOT_CONTRACT=0x0 forge script script/print_contracts.s.sol:PrintContractsScript --rpc-url $RPC_URL -vvvv
 ```
 
+And you can also run the set of RPC tests against the newly deployed pool:
+```bash
+ROOT_CONTRACT=0x0 MAKER_RPC_TESTS=false forge script script/run_rpc_tests.s.sol:RunRPCTests --rpc-url $RPC_URL
```
xmxanuel, line 58:
```suggestion
ROOT_CONTRACT=0x0 MAKER_RPC_TESTS=false forge script script/run-rpc-tests.s.sol:RunRPCTests --rpc-url $RPC_URL
```

script/rpc-tests/rpc-tests.sol
```diff
@@ -0,0 +1,321 @@
...
+        reserve = Reserve(address(lenderDeployer.reserve()));
+        clerk = Clerk(address(adapterDeployer.clerk()));
+        juniorToken = T_ERC20(address(lenderDeployer.juniorToken()));
+        seniorToken = T_ERC20(address(lenderDeployer.seniorToken()));
+        currency = T_ERC20(address(Reserve(address(lenderDeployer.reserve())).currency()));
+        registry = new Title("TEST", "TEST");
+
+        // cheat: give testContract permissions on root contract by overriding storage
+        // storage slot for permissions => keccak256(key, mapslot) (mapslot = 0)
+        vm.store(address(root), keccak256(abi.encode(address(this), uint256(0))), bytes32(uint256(1)));
```
xmxanuel, line 83: :1

xmxanuel (approved):
looks good to me.

I deployed the contracts with the script to Görli, printed the address and run the rpc-test
on the fresh deployment.

PR #628, opened 2022-11-24
NAV fixes & improvements

ilinzweilin:


test/borrower/feed/navfeedPV.t.sol
```diff
@@ -200,15 +207,41 @@ contract NFTFeedTest is DSTest, Math {
...
+        assertEq(feed.currentNAV(), borrowAmount);
+        assertEq(feed.latestNAV(), borrowAmount);
+
+        feed.writeOff(loan);
+        assertEq(pile.values_uint("changeRate_loan"), loan);
+        assertEq(pile.values_uint("changeRate_rate"), feed.WRITEOFF_RATE_GROUP());
+        assertEq(feed.currentNAV(), 0);
+        assertEq(feed.latestNAV(), 0);
+        assert(feed.zeroPV(loan));
+        feed.borrow(loan, borrowAmount); // try to borrow again
```
offerijns, line 244: Just a comment that we don't need to address now, but for the future: with Foundry, you can use `vm.expectRevert` instead of `testFail`. This way you can check that a specific revert happens, rather than that one of the calls in the entire test fails.

offerijns (approved):

src/borrower/feed/navfeedPV.sol
```diff
@@ -153,18 +162,23 @@ contract NAVFeedPV is Auth, Math {
...
-        // calculate NAV change
-        latestNAV = safeSub(nav, outstandingDebt);
+        pile.changeRate(loan, WRITEOFF_RATE_GROUP);
+        // calculate & update NAV change
+        if (nav >= outstandingDebt) {
+            latestNAV = safeSub(nav, outstandingDebt);
+        } else {
+            latestNAV = 0;
+        }
         lastNAVUpdate = block.timestamp;
```
xmxanuel, line 182:
```suggestion
  calcUpdateNAV();
```

xmxanuel (commented):
I think the write-off logic can be simplified. The case is a bit different for `borrow` and `repay`.
The rest looks good

src/borrower/feed/navfeedPV.sol
```diff
@@ -128,14 +128,23 @@ contract NAVFeedPV is Auth, Math {
...
         return amount;
     }
 
     function repay(uint256 loan, uint256 amount) external virtual auth {
-        latestNAV = safeSub(currentNAV(), amount);
-        lastNAVUpdate = block.timestamp;
+        if (pile.loanRates(loan) != WRITEOFF_RATE_GROUP) {
+            // only reduce NAV, if loan is not written off
+            uint256 nav = currentNAV();
```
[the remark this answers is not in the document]
ilinzweilin: that's what the currentNav function does, only recalculates, if no calculation happened in the same block

src/borrower/feed/navfeedPV.sol
```diff
@@ -153,18 +162,23 @@ contract NAVFeedPV is Auth, Math {
     function unlockEvent(uint256 loan) public virtual auth {}
 
     function writeOff(uint256 loan) public auth {
-        uint256 nav = currentNAV(); // retrieve latest nav
+        require(pile.loanRates(loan) != WRITEOFF_RATE_GROUP, "loan already written off");
+        uint256 nav = currentNAV();
```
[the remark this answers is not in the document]
ilinzweilin: we need the currentNAV as we are also retrieving the most recent debt of the loan. If we don't use the current NAV and there is only 1 loan -> loanDebt will be bigger than NAV

src/borrower/feed/navfeedPV.sol
```diff
@@ -177,14 +185,18 @@ contract NAVFeedPV is Auth, Math {
         if (lastNAVUpdate == block.timestamp) {
             return latestNAV;
         }
-        uint256 totalDebt;
+        return poolValue();
+    }
+
+    function poolValue() internal view returns (uint256) {
+        uint256 poolValue;
```
xmxanuel, line 192: A local variable same a function name will produce a warning

src/borrower/feed/navfeedPV.sol
```diff
@@ -128,14 +128,23 @@ contract NAVFeedPV is Auth, Math {
...
         return amount;
     }
 
     function repay(uint256 loan, uint256 amount) external virtual auth {
-        latestNAV = safeSub(currentNAV(), amount);
-        lastNAVUpdate = block.timestamp;
+        if (pile.loanRates(loan) != WRITEOFF_RATE_GROUP) {
+            // only reduce NAV, if loan is not written off
+            uint256 nav = currentNAV();
+            if (nav >= amount) {
```
xmxanuel, line 141:
```suggestion
            if (nav > amount) {
```
Saving a bit of gas for the zero case :)

xmxanuel (approved):

PR #629, opened 2022-11-25
rpc-tests: deactivate clerk for basic tests

xmxanuel:
**Basic RPC Tests**
 - remove `clerk` dependency from pool for basic rpc-test if needed
     - optional removal from assessor
     - optional removal from reserve
     
**MKR RPC Tests**
- revert immediately if MKR is not live
   - for example MKR spell not executed


offerijns (approved):
LGTM!

PR #630, opened 2022-12-01
Simple clerk

AStox:
This PR adds the simpleClerk.sol contract as a significantly simpler alternative to clerk.sol


AStox:
Closing since this was de-prioritized with the deprecation of Tinlake.

PR #633, opened 2022-12-29
Change write-off wrapper to accept root as parameter

AStox:
The current writeoff wrapper assumes the NAV feed has a public shelf and pile variable, but on the feeds this will be used on, those variables don't exist. This PR changes the writeoff wrapper to take those addresses as arguments instead

### DIFF
--- a/src/borrower/wrappers/writeOffWrapper.sol
+++ b/src/borrower/wrappers/writeOffWrapper.sol
@@ -27,7 +27,6 @@ interface FeedLike {
 }
 
 interface RootLike {
-    function relyContract(address, address) external;
     function borrowerDeployer() external view returns (address);
 }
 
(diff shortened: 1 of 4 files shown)


src/borrower/wrappers/writeOffWrapper.sol
```diff
@@ -26,6 +26,17 @@ interface FeedLike {
     function maturityDate(bytes32 nft_) external view returns (uint256);
 }
 
+interface RootLike {
+    function relyContract(address, address) external;
```
offerijns, line 30:
```suggestion
```

src/borrower/wrappers/writeOffWrapper.sol
```diff
@@ -51,13 +62,13 @@ contract WriteOffWrapper is Auth, Discounting {
     }
 
     /// @notice writes off an overdue loan
-    /// @param loan the loan id
-    /// @param nftFeed address of the feed
-    function writeOff(uint256 loan, address nftFeed) public auth {
-        FeedLike feed = FeedLike(nftFeed);
-        PileLike pile = PileLike(feed.pile());
+    /// @param root the address of the root contract
+    function writeOff(uint256 loan, address root) public {
```
offerijns: Seems more logical to me:
```suggestion
    function writeOff(address root, uint256 loan) public {
```

hieronx (dismissed):

hieronx (approved):

